PR #1, opened 2023-07-26
Check student can submit assignment + Use onyen instead of student ID

ptlharit2:
Check student can submit assignment + Use onyen instead of student ID

### DIFF
--- a/app/api/api_v1/endpoints/submission_router.py
+++ b/app/api/api_v1/endpoints/submission_router.py
@@ -19,7 +19,7 @@ def create_submission(
     assignment_id: int,
     commit_id: str
 ):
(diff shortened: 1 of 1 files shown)


frostyfan109:
@ptlharit2 Looks good, but you can't do `student_id=onyen`. `student_id` is a pkey integer, onyen is just a unique string. So this should look up student using the `onyen` to find the `student_id`.

ptlharit2:
Oh I misunderstood when we spoke, I thought we'd replace the use of student_id with the onyen. Makes sense now, i'll push an update

waTeim (approved):

waTeim:
I don't think the term "onyen" should be used at all, that makes it UNC-only.

PR #4, opened 2023-07-31
Alembic time

waTeim:
Add migrations support via alembic to manage database schema creation.

Is run just prior to starting the server as part of pod start.py


frostyfan109:
@waTeim I think `sqlalchemy.url` should be moved outside of alembic.ini, it shouldn't be hardcoded. The [generally agreed upon solution](https://stackoverflow.com/questions/22178339/is-it-possible-to-store-the-alembic-connect-string-outside-of-alembic-ini) seems to be computing it in `alembic/env.py`. I wonder if we could pull the url directly from `app/core/config.py`, so that we only have one place in the codebase where it's computed; if not, it would be fine to compute it from the same environment variables as well.

I think target_metadata should be `Base.metadata`, with `from app.database import Base`, because each model is using the same `declarative_base` instance. It's less confusing.

frostyfan109 (approved):

waTeim:
updated to use app.database.Base

PR #5, opened 2023-08-02
Feature/accommodation time

frostyfan109:
- Adds a `base_extra_time` column to Student so that professors can set a baseline extra time to give a student on all their assignments (e.g. for accommodation purposes)
- Add `join_date` and `exit_date` to Student to capture if a student is in the class or not when an assignment is due for the purposes of generating grade reports
- Change `default` -> `server_default` and enable `compare_server_default` so that alembic picks up on defaults in migrations
- Add automatic type migration to alembic, which will catch and migrate if a column type is changed
- Add uvicorn cli args to `start.py`, including reloader


waTeim:
looks like some of the same files are changed in versions as the other PR (makes sense), but I wonder what the effect will be.  If they are identical, the probably no conflict?

waTeim (approved):

PR #6, opened 2023-08-02
Feature/professor table

frostyfan109:
- Adds an `Instructor` table, removes `professor_onyen` from `Student` table in favor of this. This is to support the case where multiple instructors are teaching a class together
- Append instructors to the course metadata endpoint in the API


waTeim:
although TBH, I'd prefer `main(args)`

waTeim (approved):

waTeim (approved):

PR #7, opened 2023-08-03
404s/etc. not 400s

Hoid:
This PR changes the HTTP Status Codes used in the submission_router. Instead of using a 400 for everything, I think it makes more sense to use 404 for the cases where the resource doesn't exist and 423 (Locked) where the resource isn't available because it has been locked away from the student, the reason for which is explained in the `detail` field.

### DIFF
--- a/app/api/api_v1/endpoints/submission_router.py
+++ b/app/api/api_v1/endpoints/submission_router.py
@@ -22,13 +22,13 @@ def create_submission(
     student = db.query(StudentModel).filter_by(student_onyen=onyen).first()
     assignment = db.query(AssignmentModel).filter_by(id=assignment_id).first()
     if student is None:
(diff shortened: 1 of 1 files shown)


ptlharit2 (approved):
Agreed, these status codes are more accurate to the exception

frostyfan109:
I do think we should be using 403, not 423, for inaccessible resources. Generally anything higher than 409 is very specialized and nuanced, and tend to be misused/misunderstood.

Hoid:
> I do think we should be using 403, not 423, for inaccessible resources. Generally anything higher than 409 is very specialized and nuanced, and tend to be misused/misunderstood.

I'm okay with that, just figured that even though 423 is pretty specialized it works really well for that specific use case and makes intuitive sense to the client. But if you want to change it I'm okay with it.

PR #10, opened 2023-08-04
Add endpoint to fetch latest submission

ptlharit2:
This PR adds an endpoint to fetch the latest submission for a student & assignment.

Pending some testing before I open it up for review, but let me know if there are any glaring mistakes

### DIFF
--- a/app/api/api_v1/endpoints/submission_router.py
+++ b/app/api/api_v1/endpoints/submission_router.py
@@ -3,7 +3,7 @@ from pydantic import BaseModel
 from fastapi import APIRouter, Depends, HTTPException
 from fastapi_pagination import Page
 from fastapi_pagination.ext.sqlalchemy import paginate
-from sqlalchemy import select
+from sqlalchemy import desc
 from sqlalchemy.orm import Session
 
 from app.models import SubmissionModel, StudentModel, AssignmentModel
@@ -65,5 +65,5 @@ def get_submission(
 ):
     student = db.query(StudentModel).filter_by(student_onyen=onyen).first()
     assignment = db.query(AssignmentModel).filter_by(id=assignment_id).first()
-    submission_commit_id = assignment.get_latest_submission_commit(db, student.id)
+    submission_commit_id = db.scalar(SubmissionModel.commit_id).filter_by(student_id=student.id, assignment_id=assignment.id).order_by(desc(SubmissionModel.submission_time))
     return submission_commit_id
--- a/app/api/api_v1/endpoints/submission_router.py
+++ b/app/api/api_v1/endpoints/submission_router.py
@@ -20,6 +20,10 @@ class SubmissionBody(BaseModel):
 def validate_submission(db: Session, onyen: int, assignment_id: int):
     student = db.query(StudentModel).filter_by(student_onyen=onyen).first()
     assignment = db.query(AssignmentModel).filter_by(id=assignment_id).first()
+
+    # TODO: We should validate that the submitted commit id actually exists in gitea before persisting it in the database.
+    # We don't want another component of EduHeLx to assume the commit we return exists and crash when it doesn't.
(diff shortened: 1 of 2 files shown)


app/api/api_v1/endpoints/submission_router.py
```diff
@@ -29,13 +22,36 @@ def create_submission(
...
+
+@router.get("/submission/", response_model=Page[SubmissionSchema])
+def get_submission(
+    *,
+    db: Session = Depends(get_db),
+    onyen: int,
+    assignment_id: int
+):
+    student, assignment = validate_submission(db, onyen, assignment_id)
+    submission = db.query(SubmissionModel).filter_by(student_id=student.id, assignment_id=assignment.id, submission_time=assignment.get_latest_submission_time(db, onyen)).first()
```
Hoid, line 56:
```suggestion
    onyen: int,
    assignment_id: int
):
    student, assignment = validate_submission(db, onyen, assignment_id)
    submission = db.query(SubmissionModel)
                    .filter_by(
                        student_id=student.id, 
                        assignment_id=assignment.id, 
                        submission_time=assignment.get_latest_submission_time(db, onyen))
                    .first()
```suggestion
    onyen: int,
    assignment_id: int
):
    student, assignment = validate_submission(db, onyen, assignment_id)
    submission = db.query(SubmissionModel).filter_by(student_id=student.id, assignment_id=assignment.id, submission_time=assignment.get_latest_submission_time(db, onyen)).first()
```
```

Comment (Hoid) [inline app/api/api_v1/endpoints/submission_router.py]:
```suggestion
    submission = db.query(SubmissionModel)
                    .filter_by(
                        student_id=student.id, 
                        assignment_id=assignment.id, 
                        submission_time=assignment.get_latest_submission_time(db, onyen))
                    .first()
```

Hoid: I feel like this looks better/doesn't word wrap as much, but it still word wraps the word `onyen))`, which sucks. If we can figure out how to easily fix that then cool, but if not I think this is better. What do you think @ptlharit2 ?

frostyfan109 (changes requested):
I'll write what I sent via Slack here:

I like that validation was moved into its own function, but `get_submission` should not validate that a submissions can still be uploaded; this endpoint will only be called after the assignment has closed.

`get_submission` should not be paginated. It should just return the `commit_id` of the latest submission for a particular student/assignment.

Instead of adding a `get_latest_submission_time` method to the `Assignment` model, I would use:
```db.scalar(SubmissionModel.commit_id).filter_by(student_id=student.id, assignment_id=assignment.id).order_by(desc(SubmissionModel.submission_time))```

app/api/api_v1/endpoints/submission_router.py
ptlharit2: yup agreed, I changed the query however in my recent commit so it looks less verbose and a little better

ptlharit2:
@frostyfan109 I added a few changes from your suggestion, and I also merged the changes you pushed to `develop`. did I miss anything?

frostyfan109:
> @frostyfan109 I added a few changes from your suggestion, and I also merged the changes you pushed to `develop`. did I miss anything?

You deleted the line I added that verifies the assignment is available. It's confusing, but we need 3 checks for status, not 2. An assignment can be `released`, `available`, or `closed`.

Full rundown:
- Released means that a professor has set both an open and close date for the assignment.
- Available means that the assignment has opened for the student (i.e. the adjusted open date has passed for the student). it may also be closed for the student (which takes precedence in status).
- Closed means that the assignment has closed for the student (i.e. the adjusted due date has passed for the student).

do you have any suggestions of alternative names than `released` that we should change this to? i admit it's a confusing name.

Hoid:
> do you have any suggestions of alternative names than `released` that we should change this to? i admit it's a confusing name.

I'd suggest `created`, because it's been created but not yet set to "available".

frostyfan109 (approved):
👍

PR #11, opened 2023-08-07
Create submission endpoint changes

frostyfan109:
- Fix onyen: int -> str
- Change create submission args to json body, instead of query params
- Verify that the assignment has opened for a student
- Change 423 -> 403 error codes
  - 423 traditionally implies that another person has placed a lock on a file (e.g. someone editing a file), which isn't really what we're trying to indicate here.

### DIFF
--- a/app/api/api_v1/endpoints/submission_router.py
+++ b/app/api/api_v1/endpoints/submission_router.py
@@ -1,4 +1,5 @@
 from typing import List
+from pydantic import BaseModel
 from fastapi import APIRouter, Depends, HTTPException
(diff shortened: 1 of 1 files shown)


ptlharit2 (approved):

app/api/api_v1/endpoints/submission_router.py
```diff
@@ -1,4 +1,5 @@
 from typing import List
+from pydantic import BaseModel
```
ptlharit2, line 2: are you planning to use this later? if not, do we need this import?

PR #14, opened 2023-08-18
Add endpoint for profs to update the assignment deadline

ptlharit2:
Added an endpoint which takes the `assignment_id` and `new_due_date`, and updates the previous deadline in the DB

https://renci.atlassian.net/browse/EDU-263


app/api/api_v1/endpoints/assignment_router.py
```diff
@@ -41,4 +42,22 @@ def get_assignment_submissions(
         .filter_by(assignment_id=assignment_id) \
         .join(StudentModel) \
         .filter(StudentModel.student_onyen == onyen) \
-        .all()
\ No newline at end of file
+        .all()
+
+@router.post("/assignment/{assignment_id}/deadline", response_model=AssignmentSchema)
```
Hoid: This is more of a `put` than a `post`, because you're updating a resource within the database explicitly and not creating a new resource.

app/api/api_v1/endpoints/assignment_router.py
```diff
@@ -41,4 +42,22 @@ def get_assignment_submissions(
...
+    db: Session = Depends(get_db),
+    assignment_id: int,
+    new_due_date: str
+):
+    assignment = db.query(AssignmentModel).filter_by(id=assignment_id).first()
+    if assignment is None:
+        raise HTTPException(status_code=404, detail="Assignment does not exist")
+    
+    # convert the input param string to a datetime object
+    new_due_date = datetime.strptime(new_due_date, '%Y-%m-%d %H:%M:%S')
```
Hoid: Gotta catch the `ValueError` that's thrown if the new_due_date or format can’t be parsed. If it's thrown I would raise a new HTTPException with a 400 error code (bad request).

Hoid: I would also move this to before the db query, so we can tell the user they have the wrong input format before we even query the db.

Hoid (changes requested):

app/api/api_v1/endpoints/assignment_router.py
ptlharit2: yeah good call on validating it earlier, before the query

app/api/api_v1/endpoints/assignment_router.py
frostyfan109: You should actually be using `PATCH` since it's a partial update to the resource, use `PUT` when replacing an existing resource.

app/api/api_v1/endpoints/assignment_router.py
frostyfan109: You should be able to use `datetime.datetime` directly in the arguments to accept a timezoned datetime, and FastAPI will handle the validation and parsing.

frostyfan109:
@ptlharit2 

I think this endpoint should allow someone to modify all the relevant assignment metadata. So in the JSON body, each field is optional:
- `name: str | None`
- `directory_path: str | None`
- `available_date: datetime | None`
- `due_date: datetime | None`

Because we'll need to add functionality to modify all of these down the line. It's best practice in RESTful API design to use a single PATCH endpoint to do partial updates to resources, not a bunch of endpoints for each modification.

It's also worth pointing out that the `Assignment` model in the database has a `last_modified_date` column. We aren't using this anywhere at the moment, but e.g. it may be relevant if we add a UI for the professors. I'd like to add a mixin to the model to automatically handle tracking this, but for the moment, the patch endpoint should manually update this when it modifies assignments.

app/api/api_v1/endpoints/assignment_router.py
ptlharit2: @Hoid @frostyfan109 I changed to PATCH now in the recent commit

app/api/api_v1/endpoints/assignment_router.py
ptlharit2: using datetime now, and the validation should be handled within the json body

frostyfan109:
@ptlharit2 I'm realizing now that `available_date` and `due_date` are actually nullable columns in the database table. I'm not sure how to distinguish between a field that is None because the user wants it set to null, or None because the user omitted it in the request body since they don't want to change it and it defaulted to None. Please research some on how to do this. I think [this documentation](https://fastapi.tiangolo.com/tutorial/body-updates/#using-pydantics-exclude_unset-parameter) shows the intended way to do this.

Also, the JSON body should be `UpdateAssignmentBody`, not `Updated`

ptlharit2:
I added changes to be able to patch any field of an assignment. It also differentiates between fields that are set as None (for nullable fields in our model) and fields that aren't set at all. I addressed other feedback too

app/api/api_v1/endpoints/assignment_router.py
```diff
@@ -28,5 +37,32 @@ async def get_assignments(
         student_assignment = await StudentAssignmentService(db, student, assignment).get_student_assignment_schema()
         student_assignments.append(student_assignment)
         
+    return student_assignments
+
+@router.patch("/assignment/{assignment_name}", response_model=AssignmentSchema)
```
Hoid, line 42: why are we using `assignment_name` instead of `assignment_id` here? once the assignment is created in the db it should return an `id` that we can use everywhere, rather than a `name` that potentially has spaces and has to be encoded or whatever, imo. it will prooobably work, but working with IDs is way better than working with long strings if you can help it. what do you think @frostyfan109?

ptlharit2: we auto populate the assignment_id when creating the assignment. so unless the prof saves the ids that  we return, or unless we create a view where they can see the assignment ids, they won't know what to input for the script no?

app/api/api_v1/endpoints/assignment_router.py
```diff
@@ -28,5 +37,32 @@ async def get_assignments(
...
+    # Differentiate between fields that are set as None and fields that are not set at all
+    updated_set_fields = assignment_body.dict(exclude_unset=True)
+
+    if updated_set_fields.new_name is not None:
+        assignment.name = updated_set_fields.new_name
+    if updated_set_fields.directory_path is not None:
+        assignment.directory_path = updated_set_fields.directory_path
+
+    assignment.available_date = updated_set_fields.available_date
+    assignment.due_date = updated_set_fields.due_date
```
Hoid: Why are the first two vars behind an if check and the second two not?

app/api/api_v1/endpoints/assignment_router.py
Hoid: Oh, I was thinking there would be an endpoint that would input the assignment name and return the assignment id to make sure that the assignment exists first, but I suppose that that's more expensive than just doing things the way you're doing it. Let's keep it this way.

app/api/api_v1/endpoints/assignment_router.py
ptlharit2: the first two vars are mandatory and non nullable fields in our model. so I am checking here before I update the value in the DB.

the second two vars are allowed to be None in our model. so if the prof specifically wants to set them as None, and provide that as part of the request body, we allow that

frostyfan109:
@ptlharit2 as far as I can tell, you haven't addressed the issue where the date times here can be actual nullable values in the db. Fast api will put None as the default in the request body for any parameters that aren't passed in (which is distinct from the behavior we want; that is, the field should not be an attribute of the request body if it isn't passed. It should be None if the user wants to set it to null). You need to use that feature I linked the docs to to do this properly.

Hoid:
The only things left imo is what Griffin mentioned and unit tests.

frostyfan109:
@Hoid unit tests haven't been added yet, and I haven't had the time to devote to adding them. May want to create a ticket if we want to prioritize them.

ptlharit2:
> @ptlharit2 as far as I can tell, you haven't addressed the issue where the date times here can be actual nullable values in the db. Fast api will put None as the default in the request body for any parameters that aren't passed in (which is distinct from the behavior we want; that is, the field should not be an attribute of the request body if it isn't passed. It should be None if the user wants to set it to null). You need to use that feature I linked the docs to to do this properly.

based on the FastAPI docs the `updated_set_fields = assignment_body.dict(exclude_unset=True)` removes all unset fields instead of defaulting them to None. unless I misunderstood what you meant?

frostyfan109:
@ptlharit2 my bad I completely missed this. The dict method does return a dictionary though, so you need to change the accessing to use dict indexing or .get I think.

ptlharit2:
> @ptlharit2 my bad I completely missed this. The dict method does return a dictionary though, so you need to change the accessing to use dict indexing or .get I think.

oh shoot you are right, yeah let me push a change and do a quick test

ptlharit2:
@Hoid @frostyfan109 I added that change, is there anything else I missed?

frostyfan109:
@ptlharit2 This looks good. Merging might be a bit painful. Let me know if you want to talk about how to do the merge.

ptlharit2:
@frostyfan109 Is there a process to merge into Develop? I already fixed all the merge conflicts between our two PRs in a commit earlier

frostyfan109:
There's no process, just commit the merge so I can make sure it looks okay @ptlharit2

Hoid (approved):

PR #15, opened 2023-08-24
Auth, major refactors & improvements

frostyfan109:
- Refactor endpoint/model logic into service layer
- Add standardized exception handling
- Change student/instructor models to inherit from user model
  - Add email column, password column hashed using bcrypt, role_name column, student/instructor creation method, 
- Adds JWT/login to API
  - Tokens contain user id and user onyen
- Adds roles/permissions
  - Student/instructor/admin roles. Permissions have been assigned to the student/instructor roles as appropriate, but may need to be modified down-the-line as scope increases. New roles will likely need to get created for service accounts. The admin role is special in that it has no actual permissions assigned to it in the database. All permissions are granted to the admin role directly in the base role permission dependency in the codebase.
  - Get (list)/create/modify/delete permission for each relevant resource: student/instructor/course/assignment/submission
  - Get (list) permission on a user's own resources is implied (e.g. a student does not need the `submission:get` permission to list their own submissions
  - Roles/permissions/role-permission bindings are automatically created in the database by a migration, so there is no need to manually set them up again when setting up a new database.
  - Auth can be disabled (in development mode) by setting the environment variable `DISABLE_AUTHENTICATION=true`. This doesn't actually disable auth, but it will essentially treat everyone as an admin (even if you aren't logged in). However, only non-user-specific endpoints will be functional without logging in (e.g. if it doesn't know who you are, it can't return your submissions for an assignment).
- Change most student-specific endpoints to `/self`, now read onyen from JWT payload


alembic/versions/27622169eb7d_change_user_role_id_foreign_key_to_a_.py
```diff
@@ -0,0 +1,34 @@
...
+# revision identifiers, used by Alembic.
+revision = '27622169eb7d'
+down_revision = '6d647b5d5baa'
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('user_account', sa.Column('role_name', sa.Text(), nullable=False))
```
Hoid, line 21: I don't quite understand what this is doing. Can you help me understand?

frostyfan109: This migration is just dropping the `role_id` column and its foreign key constraint and adding in a `role_name` column with a foreign key constraint. Both are unique, but using the name is more convenient

Hoid: Oh gotcha. I was reading it wrong, thinking that it was adding a `user_account` column and then dropping it 3 lines later, but it's not. Thanks

alembic/versions/52699a7250ae_add_email_password_fields_to_student_.py
```diff
@@ -0,0 +1,30 @@
...
+revision = '52699a7250ae'
+down_revision = '6bef426b477e'
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('student', sa.Column('email', sa.Text(), nullable=False))
+    op.add_column('student', sa.Column('password', sa.Text(), nullable=False))
```
Hoid, line 22: Should we put this in another table? If we wanted to provide a student's metadata from the `students` table I don't think we should have to include the student's password too.

app/models/user/user.py
```diff
@@ -0,0 +1,28 @@
...
+    __tablename__ = "user_account"
+    
+    id = Column(Integer, Sequence("user_account_id_seq"), primary_key=True, autoincrement=True, index=True)
+    user_type = Column(Enum(UserType), nullable=False)
+
+    onyen = Column(Text, nullable=False, unique=True, index=True)
+    first_name = Column(Text, nullable=False)
+    last_name = Column(Text, nullable=False)
+    email = Column(Text, nullable=False, unique=True)
+    password = Column(Text, nullable=False)
```
Hoid: I think we should move this to another model called UserAuthModel which represents a new table called `user_auth`, which just contains username (onyen, foreign key to this user_account table) and password, plus whatever else we may put in there eventually (salt, etc.). I know this adds another table but I think it makes sense to keep the password separate from the other user metadata so that we don't have to export passwords if we want to export the student/instructor table, even if they're hashed. They are hashed right?

app/services/user_service.py
```diff
@@ -0,0 +1,108 @@
...
+
+        current_user = CurrentUser(id=user.id, onyen=user.onyen)
+
+        response = RefreshTokenSchema(
+            access_token=TokenHelper.encode(payload=current_user.dict(), expire_period=settings.ACCESS_TOKEN_EXPIRES_MINUTES),
+            refresh_token=TokenHelper.encode(payload={"sub": "refresh", **current_user.dict()}, expire_period=settings.REFRESH_TOKEN_EXPIRES_MINUTES)
+        )
+        return response
+
+class StudentService(UserService):
```
Hoid: I'd favor putting these classes in separate files, the StudentService and InstructorService.

requirements.txt
```diff
@@ -7,4 +7,6 @@ psycopg2==2.9.6
 pydantic==1.10.10
 starlette==0.27.0
 fastapi-pagination==0.12.5
-load_dotenv
+python-dotenv==1.0.0
```
Hoid, line 10: We should get this using Poetry if you're open to using it

app/schemas/course.py
```diff
@@ -1,13 +1,15 @@
...
     name: str
     master_remote_url: str
-    instructors: List[InstructorSchema]
 
     class Config:
-        orm_mode = True
\ No newline at end of file
+        orm_mode = True
+
+class CourseWithInstructorsSchema(CourseSchema):
```
Hoid, line 14: What course wouldn't have an instructor?

app/schemas/assignment.py
```diff
@@ -10,6 +10,8 @@ class AssignmentSchema(BaseModel):
     available_date: datetime | None
     due_date: datetime | None
     last_modified_date: datetime
+    # Has both an available_date and due_date set
```
Hoid, line 13: Below a few lines: what assignment wouldn't be a StudentAssignment? I feel like we could keep the schema just AssignmentSchema and move everything into there and be fine. The only thing I can think of that would be different here would be tests vs assignments.

app/core/exceptions/assignment.py
```diff
@@ -0,0 +1,21 @@
+from .base import CustomException
+
+class AssignmentNotOpenException(CustomException):
+    code = 403
+    error_code = "ASSIGNMENT_NOT_OPEN"
+    message = "assignment not open to student yet"
+
+class AssignmentClosedException(CustomException):
+    code = 403
+    error_code = "ASSIGNMENT_CLOSED"
```
Hoid: Small, but this one and the one above it don't have the double underscores after `ASSIGNMENT` like the rest do and in other files.

app/api/api_v1/endpoints/assignment_router.py
```diff
@@ -1,44 +1,32 @@
...
-from app.api.deps import get_db
+from app.schemas import StudentAssignmentSchema, AssignmentSchema
+from app.services import AssignmentService, StudentAssignmentService, StudentService
+from app.core.dependencies import get_db, PermissionDependency, UserIsStudentPermission
 
 router = APIRouter()
 
-@router.get("/assignments", response_model=List[StudentAssignmentSchema])
-def get_student_assignments(
+@router.get(
```
Hoid, line 10: If the line isn't too long (>120 chars), I'd prefer to keep these annotations to one line. What do you think?

app/api/api_v1/endpoints/auth_router.py
```diff
@@ -0,0 +1,47 @@
...
+class RefreshBody(BaseModel):
+    refresh_token: str
+    
+
+@router.post("/login", response_model=RefreshTokenSchema)
+async def login(
+    *,
+    db: Session = Depends(get_db),
+    login_body: LoginBody
+):
```
Hoid, line 24:
```suggestion
async def login(*, db: Session = Depends(get_db), login_body: LoginBody):
```
This is only 73 chars long. If it's less than 120 I think we shouldn't wrap, as it starts to actually hurt visibility instead of help.

Hoid: Same for other places below and in other files

app/api/api_v1/endpoints/student_router.py
```diff
@@ -1,24 +1,18 @@
...
 from app.schemas import StudentSchema
-from app.api.deps import get_db
+from app.services import StudentService
+from app.core.dependencies import get_db, PermissionDependency, UserIsStudentPermission
 
 router = APIRouter()
 
-@router.get("/student", response_model=StudentSchema)
-def get_student(
+@router.get("/student/self", response_model=StudentSchema)
```
Hoid, line 9: What was the inspiration for the `/self` part of the URI? I'm not sure it's necessary... I think we should just have one endpoint called `GET /student` that requires that either
1. they are a student and the onyen they're requesting is the same as their onyen, or 
2. they are a professor (who can get any student's information within their class's database)

app/api/api_v1/endpoints/submission_router.py
```diff
@@ -1,71 +1,61 @@
...
+from app.core.dependencies import get_db, PermissionDependency, UserIsStudentPermission, SubmissionCreatePermission, SubmissionListPermission
 
 router = APIRouter()
 
 class SubmissionBody(BaseModel):
-    onyen: str
     assignment_id: int
     commit_id: str
 
-def validate_submission(db: Session, onyen: str, assignment_id: int):
```
Hoid, line 20: Why did we remove this entirely? I think we still need to validate that the student has the ability to submit, and I don't see this behavior anywhere else.

I think it should just be a method on a Submission class, which contains a SubmissionBody property and this method.

Hoid (changes requested):

app/api/api_v1/endpoints/student_router.py
frostyfan109: A user can list their *owned* resources without having the explicit permission to do so. E.g. a student can list submissions they own without the SubmissionList permission. So the /self is an important distinction to indicate that the endpoint pertains to an owned resource.

I'm not in love with this solution. I could not find a straight answer on the accepted best practice here, but this seems to be a somewhat common approach.

I don't think the alternative you're suggesting is good though. It's more restful to have separate endpoints:
`GET /student/onyen` and `GET /student/self` sounds very reasonable to me. What do you think?

app/api/api_v1/endpoints/submission_router.py
frostyfan109: This wasn't removed it was just refactored into the `StudentAssignmentService` class, `SubmissionService.create_submission` uses `StudentAssignmentService.validate_student_can_submit`.

app/api/api_v1/endpoints/auth_router.py
frostyfan109: I disagree specifically for API endpoints. In general, we should be wrapping at 120 characters. However, I think that every endpoint should be wrapped regardless of length. The arguments are very contrived; you can have query params, url path params, JSON body params, form params, dependency injections, request context, etc. all being passed in through the endpoint's arguments, as well as the varargs `*` catch-all, so separating them by line makes it more clear what is actually happening in the endpoint.

app/api/api_v1/endpoints/assignment_router.py
frostyfan109: Mixed feelings about wrapping the endpoint decorators. I personally think it's a nicer structure to wrap decorator arguments on new lines when they're even somewhat lengthy (more than like 80 characters), though that's really just preference. I generally prefer to begin wrapping at fewer characters than the standard 120.

app/core/exceptions/assignment.py
frostyfan109: Good catch, will fix this.

app/schemas/course.py
frostyfan109: This schema is named as such to distinguish it from the Course schema which models the `course` table in the database. This schema doesn't represent a database table, but is used as an API response.

requirements.txt
frostyfan109: Yeah we can do that

app/services/user_service.py
frostyfan109: Will separate them

app/models/user/user.py
frostyfan109: Yes passwords are hashed and salted using bcrypt. We can technically move the onyen/password to an association  table like `UserAuth` connected using the onyen as a foreign key. I've never heard of this being done before, you may be overthinking it, but it doesn't sound unreasonable.

alembic/versions/52699a7250ae_add_email_password_fields_to_student_.py
frostyfan109: Pretty sure this is an old migration from before I added the user table. Whatever migration adds the user table changes the email/password to be in the user table, not the student table.

frostyfan109: Regardless, I'm not sure what you're getting at with the student's metadata/password thing. Why would you not store a student password in the same table as other information such as the student's name/onyen? It's not as if we are exposing this data externally, it's purely the database structure.

app/api/api_v1/endpoints/auth_router.py
Hoid: As long as we do it everywhere I guess that's fine.

app/api/api_v1/endpoints/student_router.py
Hoid: I don't see a `GET /student/{onyen}` endpoint in this file, is it somewhere else?

app/schemas/assignment.py
frostyfan109: This is similar to the CourseWithInstructor schema. The Assignment schema directly corresponds to the Assignment model in the database. So when you pull an assignment from the database, you need to add extra info to it before it's a student assignment.

frostyfan109: A regular assignment is a resource that is shared between all students. A student assignment is a constructed resource that represents said Assignment specifically for that student, and is not shared.

app/api/api_v1/endpoints/student_router.py
frostyfan109: There isn't one yet. We can add it very easily, there just hasn't been a reason to yet.

alembic/versions/52699a7250ae_add_email_password_fields_to_student_.py
Hoid: It's kind of moot now that I understand that "passwords" are autogenerated and not user-inputted. I'll mark this as resolved.

requirements.txt
frostyfan109: This doesn't belong in this PR though

.env.sample
```diff
@@ -1,5 +1,9 @@
 # Dev phase: dev | prod
 DEV_PHASE=dev
+# Disables authentication on endpoints. This option cannot be used with DEV_PHASE=prod.
+DISABLE_AUTHENTICATION=false
+# E.g. secrets.token_urlsafe(64)
```
Hoid:
```suggestion
# E.g. run secrets.token_urlsafe(64) in a python3 shell to get this value
```

Hoid (approved):

PR #17, opened 2023-09-18
Unit testing

frostyfan109:
- Adds pytest unit testing
- Mocking `settings` config module and database with mock data
- Adds fixtures for setting up an API client with user identity (logged out, admin, instructor, student)
- Adds 2 basic unit tests for demonstration
- Moves roles out of database and into their own module. Role's are still tracked by `role_name` in the `user_account` table and a custom type loads the role into the model/serializes it into the database. This should be more maintainable in the long run.


app/core/role_permissions.py
```diff
@@ -0,0 +1,73 @@
...
+    UserPermission.STUDENT__CREATE,
+    UserPermission.STUDENT__MODIFY,
+    UserPermission.INSTRUCTOR__GET,
+    UserPermission.SUBMISSION__GET
+])
+student_role = UserRole("student", [
+    UserPermission.COURSE__GET,
+    UserPermission.SUBMISSION__CREATE,
+    UserPermission.INSTRUCTOR__GET
+])
```
ptlharit2, line 72: do we not need to give students `UserPermission.SUBMISSION__GET`?

I assumed we'd allow students to fetch their own submission. And obv we only allow them to get their own submission

app/core/role_permissions.py
```diff
@@ -0,0 +1,73 @@
...
+    UserPermission.ASSIGNMENT__GET,
+    UserPermission.ASSIGNMENT__CREATE,
+    UserPermission.ASSIGNMENT__MODIFY,
+    UserPermission.ASSIGNMENT__DELETE,
+    UserPermission.COURSE__GET,
+    UserPermission.COURSE__CREATE,
+    UserPermission.COURSE__MODIFY,
+    UserPermission.STUDENT__GET,
+    UserPermission.STUDENT__CREATE,
+    UserPermission.STUDENT__MODIFY,
```
ptlharit2, line 64: should we give instructors permission to delete students for a course? on the event they drop it early in the semester

app/services/jwt_service.py
```diff
@@ -5,6 +5,8 @@
 
 class JwtService:
     async def verify_token(self, token: str) -> None:
+        print(settings.JWT_SECRET_KEY)
```
ptlharit2: do we need this print?

ptlharit2 (approved):
Looks great, left a couple minor comments

app/core/role_permissions.py
frostyfan109: Yeah, this is an understandable confusion point I had to explain to Tyler when I initially implemented it in the database. In this PR I just moved permissions out of the database because it didn't make much sense to store them there anyways.

It is implied that you can access any resource you own whether or not you explicitly have that permission. So a student doesn't need the StudentGet permission to get information about themselves. They don't need the SubmissionGet permission to get their own submissions.

Rather, the SubmissionGet permission would allow a student to list _other_ students' submissions, which is not what we want.

app/core/role_permissions.py
frostyfan109: I thought about that, but decided that we shouldn't. We have that column `exit_date` to mark that a student has dropped a course, and so I don't think we ever want to outright delete the student from the table entirely.

app/services/jwt_service.py
frostyfan109: good catch, no that should not be there. I'll remove it

frostyfan109: apparently that method isn't actually used anywhere though. I'm not exactly sure what I was debugging with it lol..

PR #18, opened 2023-09-20
Setup wizard scripts, user creation, & gitea

frostyfan109:
Adds API setup wizard scripts
- Create course and class master repo in Gitea.
- Create instructors and add them to the instructor organization/team in Gitea so they can modify the master repo.

Adds full user creation and user endpoints:
- Student/instructor creation endpoints will now store the autogen password credentials in a secret. They will also create an account in Gitea with the onyen/autogen password.
- Student creation will automatically create a fork of the master repository for the student.
- Instead of getting the student or instructor you're logged in as, it's been generalized to getting the "user" you're logged in as. This will automatically resolve the user-type of whatever the user (e.g. if you're a student, the API will return your user as a Student with all associated student fields, instead of returning a User and omitting the student-associated fields).
- May change GET student/{onyen} and GET instructor/{onyen} to a GET user/{onyen} in the future as well.

Note that class roster import feature is not done yet. But this will go into the jupyter-lab-prof repository, not the grader api.


Hoid:
@frostyfan109 can you pull develop into this branch, since you merged #17 ? I think that will get rid of some of the files in here. (at the very least it will update it with the changes you made to the PR after you pulled in the branch originally)

Hoid:
Nevermind, I went ahead and did it, there were no merge conflicts

app/api/api_v1/endpoints/instructor_router.py
```diff
@@ -0,0 +1,49 @@
...
+from sqlalchemy.orm import Session
+from app.schemas import InstructorSchema
+from app.models.user import UserType
+from app.services import InstructorService, CourseService, KubernetesService
+from app.core.utils.auth_helper import PasswordHelper
+from app.core.dependencies import get_db, PermissionDependency, InstructorListPermission, InstructorCreatePermission
+
+router = APIRouter()
+
+class CreateInstructorWithoutPasswordBody(BaseModel):
```
Hoid: Why is "WithoutPasswordBody" necessary?

frostyfan109: Just being overly opaque in my naming. No reason it's necessary per se.

Hoid: I think we should remove it, and just have it be `CreateInstructorBody`. Same for `CreateStudentWithoutPasswordBody`.

app/api/api_v1/endpoints/instructor_router.py
```diff
@@ -0,0 +1,49 @@
...
+async def list_instructor(
+    *,
+    request: Request,
+    db: Session = Depends(get_db),
+    perm: None = Depends(PermissionDependency(InstructorListPermission))
+):
+    instructors = await InstructorService(db).list_instructors()
+    return instructors
+
+@router.post("/instructor", response_model=InstructorSchema)
```
Hoid: Might be small, but I think we should use the plural as the root everywhere, e.g. `/instructors`. This also goes for line 19 as well as the student endpoints. When you're doing a POST on `/instructor` I don't think that makes much sense, since there's no resource for a single `instructor` without more information (like query params or path params), only a list of instructors. And when you do a GET it should be `/instructors/{onyen:str}` because the instructor's onyen is a subset of the `instructors`. Only reason I bring it up is because this is when we're creating these endpoints, and if we create them this way it's much harder to change them later.

app/api/api_v1/endpoints/instructor_router.py
frostyfan109: How come?

app/api/api_v1/endpoints/instructor_router.py
frostyfan109: @Hoid Right now, we use the plural of the resource name to refer to a collection of the resource. We use the singular to refer to an individual instance of the resource. There doesn't seem to be a completely agreed upon naming scheme for REST APIs. Some people like all resource names to be plural everywhere, some people like it how we have it right now.

Resource creation should definitely be plural, like POST /instructors. I'll fix this, it's a mistake. In general, do you want the API to be all plural?

Hoid: I think when you're dealing with lists of resources then yes, it should be all plural, but there are times when endpoints will be actions for the server to take rather than resources to return. In the case of an action it's likely not going to be plural. 

But for these endpoints I think endpoints underneath `/instructors` would pertain to those `instructors` (and the same goes for `/students`. It doesn't make sense to me that you'd create an instructor at `POST /instructors` (which is right) and then go off of that subpath into a new path (`/instructor`) to query that same instructor you just created. I would think that all resources and actions that pertain to instructors would be underneath the `/instructors` path, so that we'd have nested paths, like how `/instructors` gives way to `/instructors/{onyen}`, and could potentially allow for something like `/instructors/{onyen}/profile` or `/instructors/{onyen}/class-rules` in the future.

app/api/api_v1/endpoints/instructor_router.py
Hoid: The `WithoutPasswordBody` just seems superfluous, no? I don't see a strong use for it being there, and it makes the name much longer. It also implies some importance for there being no password body, and I don't see any.

frostyfan109: That's fair, I'll change this.

app/api/api_v1/endpoints/instructor_router.py
frostyfan109: I'll fix the `post /instructors` typo and make sure that the rest of our endpoints are consistent with what you've described.

app/api/api_v1/endpoints/instructor_router.py
```diff
@@ -0,0 +1,49 @@
+from typing import List
+from pydantic import BaseModel
+from fastapi import APIRouter, Request, Depends
+from sqlalchemy.orm import Session
+from app.schemas import InstructorSchema
+from app.models.user import UserType
```
ptlharit2: We are importing `from app.models.user import UserType` In many files.
But its only being used in the `kubernetes_service`, so would be cleaner if we removed the import where its not needed

app/services/course_service.py
```diff
@@ -1,18 +1,65 @@
...
 class CourseService:
     def __init__(self, session: Session):
         self.session = session
     
     async def get_course(self) -> CourseModel:
-        return self.session.query(CourseModel).first()
+        try:
+            return self.session.query(CourseModel).one()
+        except MultipleResultsFound as e:
+            raise MultipleCoursesExistException()
```
ptlharit2, line 17: If multiple courses get created, there is no possible way to fetch one course via the API with this. I'd be wary of that, because I think we also dont have a way for instructors to delete a course if they make one by mistake

app/core/exceptions/course.py
```diff
@@ -0,0 +1,16 @@
...
+    error_code = "COURSE__MULTIPLE_COURSES_EXIST"
+    message = "database misconfigured, multiple courses exist in course table"
+
+class NoCourseExistsException(CustomException):
+    code = 500
+    error_code = "COURSE__NO_COURSE_EXISTS"
+    message = "database misconfigured, no course exists in course table"
+
+class CourseAlreadyExistsException(CustomException):
+    code = 403
```
ptlharit2: minor, but I dont think 403 is the one we need. 409 might be more appropriate, or 400 -

```
400 Bad Request - when the server will not process a request because it's obvious client fault
409 Conflict - if the server will not process a request, but the reason for that is not the client's fault
```

ptlharit2: ah I just saw you used 409 for `UserAlreadyExists`, we could be consistent if we used 409 here too

ptlharit2 (commented):
The PR looks great, code speaks for itself and its easy to follow. you may have a couple non-major merge conflicts in start.py from the logger PR. I left a few comments/questions

app/services/course_service.py
frostyfan109: By design, there should always be exactly one course in the database. If there are 0, it will break other components because we consider this assumption to be true. We could go ahead and change this and add special handling everywhere that we've assumed this, but it's reasonable enough to just create a template of the course before standing up the API, and then the Professor can modify that course instead of deleting it and creating a new one.

I'm not totally opposed to allowing multiple courses, so the instructor can create a new one then delete the old one. Or just set one to "active". However, I don't see that use case being relevant at all.

frostyfan109: Also to be clear, professors don't have the permission to create/delete courses, so there's no risk involved there unless we give them that permission.

app/core/exceptions/course.py
frostyfan109: Yeah this should not be a 403. I'm honestly in favor of 400 here and changing the other one as well. 409 is most appropriate, although I think it's generally better to avoid the specialized codes since they are very rarely actually used and harder to recognize.

Hoid: I personally think there's some utility to using error codes like 409. I think there's some error codes that are really specific and that we won't get any use out of in our applications (e.g. 410 "Gone", which is to indicate that the resource is no longer present and won't be coming back, and that clients should purge it from their caches). But others, like 409 are useful in this scenario because they keep us from diluting 400. If we just use 200 for every success and 400 for every client failure and 500 for every server failure we're diluting the pool of what those statuses mean. I just feel like being more specific is better when it makes sense so that clients don't have to infer what the error is, and I think 409 makes sense here.

frostyfan109: We can use 409. Although the benefit of differentiation here is more or less lost since we assign unique codes to all our errors anyways.

app/api/api_v1/endpoints/instructor_router.py
frostyfan109: I'll go through and cleanup the unused imports

ptlharit2 (approved):

PR #19, opened 2023-09-28
Initial draft to add logger and a middleware to capture request context

ptlharit2:
Jira item: https://renci.atlassian.net/browse/EDU-271

Adding a logging middleware to log requests/responses for each API call.

Custom logger that is used - https://github.com/helxplatform/eduhelx-utils/pull/1 (Also needs review)


ptlharit2:
In hindsight, I should probably also add a unit test for the behavior when logging one of our custom exceptions

frostyfan109 (approved):
Looks good 👍

PR #24, opened 2023-11-01
Update schema again, include types

ptlharit2:


frostyfan109:
What tool are you using to generate these? It's very strange that user_role, which has been deleted for a while now, is showing up here, while the other deleted permission-related tables are gone. Also weird that the relations are just pointing to the table, not the foreign key column that they join on.

Anyways, this needs to be updated since it says that the user_role table exists (and has columns), when it actually doesn't exist anymore.

ptlharit2:
I'm using DBSchema which is the db manager I am using, it generates this diagram based on the metadata I believe. I can also edit this diagram if there were specific things to change

ptlharit2:
> What tool are you using to generate these? It's very strange that user_role, which has been deleted for a while now, is showing up here, while the other deleted permission-related tables are gone. Also weird that the relations are just pointing to the table, not the foreign key column that they join on.
> 
> Anyways, this needs to be updated since it says that the user_role table exists (and has columns), when it actually doesn't exist anymore.

yup I can change that, anything else to change?

Hoid:
Yea I'd much rather the tool point directly to the foreign key column that it references, rather than just the table itself, since that's not nearly as instructive or clear. Is there a different tool we could use? If we can't find anything then this seems fine, but if we can then I think we should.

frostyfan109:
That's the only big problem, the only other thing I see is that it doesn't annotate that a column is nullable. Or in general, doesn't visualize constraints.

ptlharit2:
@frostyfan109 @Hoid I pushed an updated diagram. I removed the `user_role` table and it now also shows whether the attributes are nullable or not.

The only thing I couldn't add with this tool is pointing to specific key columns. What do you think, is this good to merge or should I explore another tool?

frostyfan109 (approved):

frostyfan109:
Ugly but practical

PR #26, opened 2024-01-26
Service Layer Unit Tests

ptlharit2:
- We are mocking the Database queries to return expected results
- You can find the mock data being used under tests/data/database/
- These unit tests cover both the successful and unsuccessful paths
- Subject to change as I add more tested classes and remove the database mocking configuration


tests/services/assignment_service_test.py
```diff
@@ -0,0 +1,82 @@
+import unittest
+from unittest.mock import MagicMock
+from sqlalchemy.orm import Session
+from app.services import AssignmentService
+from app.models import AssignmentModel
+from app.core.exceptions import AssignmentNotFoundException
+from ..data.database.assignment import assignment_data
+
+class TestAssignmentService(unittest.IsolatedAsyncioTestCase):
```
Hoid: This file should have the same name as the class

ptlharit2: The file name is `assignment_service.py`, do we want to start prefixing with "test" followed by the class name?
Ex: `test_assignment_service.py` - rather than suffix which I have currently

Hoid: Considering what Griffin said the other day about test files always starting with "test", I would say yes. They don't have to always say test_<filename>, it could be test_<something general> though.

tests/services/assignment_service_test.py
```diff
@@ -0,0 +1,82 @@
...
+    async def asyncSetUp(self):
+        self.mock_session = MagicMock()
+        self.assignment_service = AssignmentService(session=self.mock_session)
+
+
+    async def test_get_assignment_by_id_success(self):
+        mock_assignment = assignment_data["available"]
+        self.mock_session.query().first.return_value = mock_assignment
+
+        result = await self.assignment_service.get_assignment_by_id(id=1)
```
Hoid: I don't like this paradigm of using outside data (from `assignment_data` above) and basically depending on it being of a certain state (some state referenced by id=1) for the test to work. This encourages us to use that state for multiple tests, which then causes problems if you change the state later because it could break certain tests but not others. I think each set of tests (so, each file/test class) needs to set up its own data for itself. That's the way I've always seen it done.

tests/services/test_assignment_service.py
```diff
@@ -0,0 +1,82 @@
...
+        self.mock_session.commit.return_value = None
+
+        result = await self.assignment_service.update_assignment_directory_path(assignment=mock_assignment, directory_path="new/directory/path")
+        self.assertEqual(result.directory_path, "new/directory/path")
+
+    async def test_update_assignment_available_date_success(self):
+        mock_assignment = assignment_data["available"]
+        self.mock_session.commit.return_value = None
+
+        result = await self.assignment_service.update_assignment_available_date(assignment=mock_assignment, available_date=mock_assignment.available_date)
```
Hoid: This one I think you're just using the same `available_date` from the `mock_assignment` object you started with, so it's not going to change it at all, and there's nothing being tested here. You'd have to change the available date to something else and test that it changes to that other value.

tests/services/test_assignment_service.py
```diff
@@ -0,0 +1,82 @@
...
+        self.assertEqual(result.available_date, mock_assignment.available_date)
+
+    async def test_update_assignment_due_date_success(self):
+        mock_assignment = assignment_data["available"]
+        self.mock_session.commit.return_value = None
+
+        result = await self.assignment_service.update_assignment_due_date(assignment=mock_assignment, due_date=mock_assignment.due_date)
```
Hoid: Same thing here

tests/services/test_course_service.py
```diff
@@ -0,0 +1,94 @@
...
+        result = await self.course_service.get_course_with_instructors_schema()
+        expected_result = CourseWithInstructorsSchema.from_orm(expected_course)
+
+        self.assertEqual(result, expected_result)
+
+    async def test_get_instructor_gitea_organization_name(self):
+        self.mock_session.query().one.return_value = self.mock_course
+
+        result = await self.course_service.get_instructor_gitea_organization_name()
+        assert result == "COMP 555-instructors"
```
Hoid, line 65: I tend not to like "magic numbers" in code, which "COMP 555" essentially is here since it's not clear where that comes from. If it's easy to just create a var at the top like `course_name = "COMP 555"` and use `course_name` wherever you need "COMP 555" that would be ideal, but it's not too big of a deal. It's not thaaat likely that someone will want to change this value in the future, so 🤷

Hoid: Yea on second thought I don't think this is necessary at all

cnbennett3 (approved):

tests/services/test_jwt_service.py
```diff
@@ -0,0 +1,34 @@
...
+
+class TestJwtService(unittest.IsolatedAsyncioTestCase):
+    async def asyncSetUp(self):
+        self.jwt_service = JwtService()
+
+    @patch('app.services.jwt_service.TokenHelper')
+    async def test_verify_token_valid(self, mock_token_helper):
+        mock_token_helper.decode.return_value = {"sub": "refresh", "id": 1, "onyen": "test_user"}
+        await self.jwt_service.verify_token("valid_token")
+        mock_token_helper.decode.assert_called_once_with(token="valid_token")
```
Hoid, line 14: I don't think this is testing anything useful tbh. And it will just give us coverage without utility, so we won't think to actually test it later.

tests/services/test_jwt_service.py
```diff
@@ -0,0 +1,34 @@
...
+            await self.jwt_service.verify_token("invalid_token")
+
+    @patch('app.services.jwt_service.TokenHelper')
+    async def test_refresh_access_token(self, mock_token_helper):
+        mock_token_helper.decode.return_value = {"sub": "refresh", "id": 1, "onyen": "test_user"}
+        mock_token_helper.encode.return_value = "new_access_token"
+
+        result = await self.jwt_service.refresh_access_token("refresh_token")
+
+        mock_token_helper.decode.assert_called_once_with(token="refresh_token")
```
Hoid, line 29: I don't think this `assert_called_once_with()` statement is doing much. I think the next one is useful because you're testing that the method gets called with a specific payload which is modified from the input, but this is just checking that `decode()` is called with `refresh_token`, which I don't think is that useful. Plus, we might want to one day change the way the `decode()` method works, and we don't want the tests to fail just because we changed the black box inside `refresh_access_token()` imo.

ptlharit2: The refresh token would not be accurate unless we ran the decode method on it. It's a simple test, but it serves the function of making sure we are decoding the token before we validate & throw the exception. I agree with your previous comment however that the last unit test already covers the first one. So I can remove `test_verify_token_valid` given it's redundant. The decode method is coming from the utility so I doubt that would impact our code unless they deprecate, which is a separate issue

Hoid: Ah okay, I agree, let's keep this one as is

PR #28, opened 2024-02-05
Allow .env.local file for local development

Hoid:
As of this PR, devs can create a file called `.env.local` as a copy of `.env.sample` in the root of the project and set values for all of the environment variables that need to be set. Previously, we had to change the config.py file manually to have the default values during development and make sure not to upload them to source control which was a major hassle. This will become even more useful once Harit merges his testing PR, since it fixes this issue in the Testing pane of vscode. 

The way I have it, it shouldn't have any effect if the `.env.local` file is not there, but we should keep an eye on it once we deploy this to a hosted environment with environment variables like Kubernetes for the first time.

### DIFF
--- a/app/core/config.py
+++ b/app/core/config.py
@@ -1,6 +1,7 @@
 from typing import Optional, Any, Dict, List
 from enum import Enum
 from pydantic import BaseModel, BaseSettings, PostgresDsn, Json, validator, root_validator
+import os.path
 
 class DevPhase(str, Enum):
     DEV = "dev"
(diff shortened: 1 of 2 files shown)


frostyfan109 (approved):

frostyfan109:
@Hoid I think start.py already loads `.env`, so maybe do it there instead of in settings?

ptlharit2 (approved):

app/core/config.py
```diff
@@ -46,8 +54,8 @@ class Settings(BaseSettings):
     LDAP_TIMEOUT_SECONDS: int = 5
 
     # Database
-    POSTGRES_HOST: str
-    POSTGRES_PORT: str = "5432"
```
frostyfan109, line 50: Why was the default value removed here?

frostyfan109 (changes requested):
Need to update README about how to use .env.local

PR #33, opened 2024-03-29
Sync our database with the current state of canvas

ptlharit2:
EDIT: This PR was merged into another branch to resolve merge conflicts, which was then merged into `develop` - https://github.com/helxplatform/eduhelx-grader-api/pull/37
All changes from this branch have been merged into `develop`, hence I am closing this PR.

Changes:

1. Added new canvas_service: has the logic to call the canvas API to fetch course, student, and assignment info
2. Added new lms_sync_service: has the logic to pull the canvas state, and update our database
3. Edited assignment_service, course_service, course_service, and user services to add methods needed for canvas & database diff checks
4. Added alembic migrations for database changes needed to accomadate canvas related data
5. Added new router lms_router: contains endpoints to trigger the downsync/upsync of changes from Canvas & Eduhelx
6. Added LMS based exceptions to have description exceptions on errors
7. Added service method and endpoint to upload grades for students based on grades CSV file
8. Added test class to test the Canvas integration given the API key


app/core/exceptions/assignment.py
```diff
@@ -18,4 +18,9 @@ class AssignmentNotCreatedException(CustomException):
 class AssignmentNotFoundException(CustomException):
     code = 404
     error_code = "ASSIGNMENT__NOT_FOUND"
-    message = "assignment not found"
\ No newline at end of file
+    message = "assignment not found"
+
+class NoAssignmentFetchedException(CustomException):
```
ptlharit2: code should be 404 in hindsight

app/core/exceptions/course.py
```diff
@@ -10,6 +10,11 @@ class NoCourseExistsException(CustomException):
     error_code = "COURSE__NO_COURSE_EXISTS"
     message = "database misconfigured, no course exists in course table"
 
+class NoCourseFetchedException(CustomException):
+    code = 500
```
ptlharit2: code should be 404 in hindsight

app/api/api_v1/endpoints/lms_router.py
frostyfan109, line 1: Probably just want a single "downsync" endpoint. Course id shouldn't be inferred, should be in course table. Permission can just be UserIsInstructor

app/core/exceptions/user.py
frostyfan109, line 1: This should indicate that it is specifically an LMS-related error. You could just move all the LMS exceptions into their own file and name them e.g. LMSNoStudentsFetchedException

app/models/course.py
frostyfan109, line 1: total students should be computed from the students table, not stored in the course table, unless there's something that makes this necessary

app/services/assignment_service.py
```diff
@@ -14,6 +14,28 @@
 class AssignmentService:
     def __init__(self, session: Session):
         self.session = session
+        
+    async def create_assignment(
```
frostyfan109, line 18: This should also create the folder in gitea (remove the directory_path argument, should just be the same as assignment name), the empty notebook file, and a gitignore containing:
- `*grades.csv` (we'll probably keep these in-memory if possible or in a temp file, but just in case)
- `{assignment_name}-dist` folder
- `{assignment_name}-prof.ipynb` <-- this is the master notebook that the professor puts tests into and such
- `{assignment_name}-student.ipynb` <-- this is the notebook generated by otter assign

app/services/canvas-test.py
frostyfan109, line 1: There's probably a more appropriate place to put this, maybe a bin or tests/bin folder

app/services/canvas_service.py
```diff
@@ -0,0 +1,86 @@
+from fastapi import HTTPException
+import requests
+from app.core.exceptions import NoCourseFetchedException, NoAssignmentFetchedException, NoStudentsFetchedException
+
+
+# Assuming we'll store our Canvas LMS API key securely, for now I am just hard coding it
+CANVAS_API_KEY = "YOUR_API_KEY"
+CANVAS_API_URL = "https://uncch.instructure.com/api/v1"
```
frostyfan109: Both of these should be settings

app/services/canvas_service.py
```diff
@@ -0,0 +1,86 @@
...
+
+class CanvasService:
+    def __init__(self, course_id):
+        self.session = requests.Session()
```
frostyfan109, line 15: Should use httpx to be consistent with the other API wrapper services. They use very similar APIs, but httpx also has much better native support for async operations.

app/services/canvas_service.py
```diff
@@ -0,0 +1,86 @@
...
+            raise HTTPException(status_code=500, detail=str(e)
+        )
+
+    async def get_course(self, additional_params=None):
+        try:
+            url = f"{CANVAS_API_URL}/courses/{self.course_id}"
+            
+            if additional_params:
+                # Append additional parameters to the URL
+                url += "?" + "&".join([f"{key}={value}" for key, value in additional_params.items()])
```
frostyfan109, line 39: You can just use the `params` argument to pass a dict for both `requests` and `httpx` libraries.

app/services/course_service.py
```diff
@@ -46,7 +48,10 @@ async def create_course(self, name: str) -> CourseModel:
 
         course = CourseModel(
             name=name,
-            master_remote_url=master_remote_url
+            master_remote_url=master_remote_url,
+            start_at=start_at,
```
frostyfan109, line 52: start_at/end_at/total_students should all be set during the initial downsync. you can change the setup wizard to use the canvas service to get these values in order to create the course, or just set them to null or  like 0000-00-00 and 9999-12-31.

app/services/course_service.py
```diff
@@ -60,4 +65,18 @@ async def get_instructor_gitea_organization_name(self) -> str:
     
     async def get_master_repository_name(self) -> str:
         course = await self.get_course()
-        return f"{ course.name }-class-master-repo"
\ No newline at end of file
+        return f"{ course.name }-class-master-repo"
+    
+    async def update_course(self, name: str, total_students: int, start_at: datetime, end_at: datetime) -> CourseModel:
+        print('bar')
```
frostyfan109: remove this

app/services/course_service.py
```diff
@@ -60,4 +65,18 @@ async def get_instructor_gitea_organization_name(self) -> str:
...
+        try:
+            course = await self.get_course()
+            course.name = name
+            course.total_students = total_students
+            course.start_at = start_at
+            course.end_at = end_at
+            self.session.commit()
+            return course
+        except NoCourseExistsException:
```
frostyfan109: should be raising the existing exception, not a new one. but this block appears to be redundant, no?

app/services/lms_sync_service.py
```diff
@@ -0,0 +1,138 @@
...
+        self.course_service = CourseService(session)
+        self.assignment_service = AssignmentService(session)
+        self.student_service = StudentService(session)
+        self.session = session
+
+    async def sync_course(self):
+        try:
+            canvas_course = await self.canvas_service.get_course({"include[]": "total_students"})
+
+            # Remove this after deciding what to do about null start/end dates
```
frostyfan109: Just make the start and end dates nullable? And set them to None.

app/services/lms_sync_service.py
```diff
@@ -0,0 +1,138 @@
...
+        # Delete assignments that are in the database but not in Canvas
+        for assignment in db_assignments:
+            if assignment.id not in [a['id'] for a in canvas_assignments]:
+                await self.assignment_service.delete_assignment(assignment.id)
+
+        for assignment in canvas_assignments:
+            try:
+                if(await self.assignment_service.get_assignment_by_id(assignment['id'])):
+                    #update the existing assignment
+                    await self.assignment_service.update_assignment_name(
```
frostyfan109: These methods take an assignment argument, not assignment_id.

app/services/lms_sync_service.py
```diff
@@ -0,0 +1,138 @@
...
+                        assignment_id=assignment['id'], 
+                        new_name=assignment['name']
+                    )
+
+                    await self.assignment_service.update_assignment_available_date(
```
frostyfan109: no assignment_id argument

app/services/lms_sync_service.py
```diff
@@ -0,0 +1,138 @@
...
+                        assignment_id=assignment['id'],
+                        available_date=assignment['unlock_at']
+                    )
+
+                    await self.assignment_service.update_assignment_due_date(
```
frostyfan109: no assignment_id argument

app/services/lms_sync_service.py
frostyfan109, line 1: ideally this whole service would use another service instance implementing some LMSService as an argument and use that to get data from the LMS and compare state. but that can wait I guess

app/services/lms_sync_service.py
```diff
@@ -0,0 +1,138 @@
...
+                await self.student_service.create_student(
+                    onyen=student['sis_user_id'],
+                    name=student['name'],
+                    email=student['email']
+                )
+
+        return canvas_students
+    
+    async def downsync(self):
+        self.sync_course()
```
frostyfan109: These are coroutines so have to be awaited

frostyfan109: You can use asyncio.gather to pool them together instead of running them serially

app/services/lms_sync_service.py
```diff
@@ -0,0 +1,138 @@
...
+        self.sync_assignments()
+        self.sync_students()
+
+
+# Currently only for testing purposes: a script that can be run to sync the LMS with the database
+from app.database import SessionLocal
```
frostyfan109: Remove this code

frostyfan109: Also I get this is for debugging but let me just leave here it's very important to always call sess.close() whenever you open a session using SessionLocal() or the connection pool can fill causing errors.

app/core/exceptions/user.py
ptlharit2: agreed, change incoming

app/models/course.py
ptlharit2: do you need the # of total students as easy access? to be able to compute the % of students that have graded submissions?

Or can you infer that from GET `/students`

ptlharit2: regardless, I agree the course table is the wrong place for this

frostyfan109: Yes basically I'm saying that you can just compute total students from the number of rows in the student table. We don't need to have an actual column tracking that

ptlharit2: yup I added a method in the student service to get the total # from the student table

app/services/lms_sync_service.py
ptlharit2: thanks, I changed it

frostyfan109: no like these methods don't take the assignment_id argument. They take the `assignment` argument which is an AssignmentModel. you can change that to an assignment_id if you really want, regardless it'll need to invoke the assignment service somewhere to actually get the model to update.

frostyfan109: you need to change the calls in the modify endpoint as well

app/services/course_service.py
frostyfan109: also a feature I recently learned of when raising a different exception from an existing one is that you can do `raise from`
```
except Exception as exception:
    raise NewException from exception
```
though I'm not sure exactly how it works or if it would work for our exceptions. requires some testing

app/services/lms_sync_service.py
ptlharit2: yeah good refactor, added to next commit

app/services/course_service.py
ptlharit2: yeah agreed, initially I had added it to catch the exception in the case of creating a course for the first time. But I am now handling that logic in `lms_sync_script`. Removing this exception from the course service

app/services/canvas-test.py
ptlharit2: yup should be under the test folder, moving it over in the next commit

app/api/api_v1/endpoints/lms_router.py
ptlharit2: added a single `downsync` endpoint to handle it all. but I left the other endpoints, we can always delete later if we realize we never want to do a partial sync

app/services/canvas-test.py
ptlharit2: done

app/services/canvas_service.py
ptlharit2: agreed with you to do add consistency in the next PR

app/services/canvas_service.py
ptlharit2: as agreed to add this in the next work PR

app/services/lms_sync_service.py
ptlharit2: That may limit our options - if we allow dates to be nullable, we are limiting out ability to date order assignments, check which need to be made unavailable, etc. Not saying we shouldn't make them nullable, just there are trade-offs. For now I can add default values

frostyfan109: You can still order any assignments which do have their dates set; even if you set the dates to 00-00-0000 and 12-31-9999, those assignments are going to have to be ordered the same way as if they had a null date, there is no difference (e.g. you would be ordering vs their "relative due date" that you set way into the future or in the null scenario just vs assignment creation date)

frostyfan109: @ptlharit2 What do you mean by "check which need to be made unavailable"? And how do dates being nullable affect that?

app/api/api_v1/endpoints/assignment_router.py
```diff
@@ -16,35 +16,13 @@ class UpdateAssignmentBody(BaseModel):
...
-    return student_assignments
-
 @router.patch("/assignments/{assignment_name}", response_model=AssignmentSchema)
 async def update_assignment_fields(
     *,
     db: Session = Depends(get_db),
     assignment_name: str,
-    assignment_body: UpdateAssignmentBody
+    assignment_body: UpdateAssignmentBody,
+    perm: None = Depends(PermissionDependency(UserIsInstructorPermission))
```
frostyfan109: should be `AssignmentModifyPermission` technically

app/api/api_v1/endpoints/assignment_router.py
```diff
@@ -63,5 +41,31 @@ async def update_assignment_fields(
         assignment = await AssignmentService(db).update_assignment_available_date(assignment, updated_set_fields["available_date"])
     if "due_date" in updated_set_fields:
         assignment = await AssignmentService(db).update_assignment_due_date(assignment, updated_set_fields["due_date"])
+    if "available_date" in updated_set_fields or "due_date" in updated_set_fields:
```
frostyfan109: Shouldn't we be upsyncing name changes as well? And this block should be separate from the other if's, its proximity implies it is performing the same function when really it is not. Maybe upsync should just always be called?

app/services/assignment_service.py
```diff
@@ -14,6 +14,51 @@
...
+        self.session.commit()
+
+        assignment_folder = f"{ name }_{ id }-dist"
+        master_notebook = f"{ name }-prof.ipynb"
+        master_repository_name = await course_service.get_master_repository_name()
+        owner = await course_service.get_instructor_gitea_organization_name()
+
+        files_to_modify = [
+            FileOperation(content="", path=f"{ assignment_folder }/{ master_notebook }", operation=FileOperationType.CREATE),
+            FileOperation(content=f"*grades.csv \n { master_notebook }", path=f"{ assignment_folder }/.gitignore", operation=FileOperationType.CREATE),
```
frostyfan109: shouldn't be spaces between `.csv\n{master_notebook}`

app/services/assignment_service.py
```diff
@@ -14,6 +14,51 @@
...
+        self.session.add(assignment)
+        self.session.commit()
+
+        assignment_folder = f"{ name }_{ id }-dist"
+        master_notebook = f"{ name }-prof.ipynb"
+        master_repository_name = await course_service.get_master_repository_name()
+        owner = await course_service.get_instructor_gitea_organization_name()
+
+        files_to_modify = [
+            FileOperation(content="", path=f"{ assignment_folder }/{ master_notebook }", operation=FileOperationType.CREATE),
```
frostyfan109: ipynb are json so an empty notebook isn't actually an empty file, should really be:
```
{
 "cells": [],
 "metadata": {},
 "nbformat": 4,
 "nbformat_minor": 5
}
```

app/services/assignment_service.py
```diff
@@ -14,6 +14,51 @@
...
+            FileOperation(content=f"*grades.csv \n { master_notebook }", path=f"{ assignment_folder }/.gitignore", operation=FileOperationType.CREATE),
+        ]
+
+        await gitea_service.modify_repository_files(
+            name=master_repository_name,
+            owner=owner,
+            branch_name="master",
+            commit_message=f"Create assignment { name }",
```
frostyfan109: probably just "Initialize assignment" is fine. You should add a method to CourseService similar to get_master_repository_name for `get_main_branch_name`. should call a static method, which can literally just return "master"

app/services/assignment_service.py
```diff
@@ -58,6 +103,31 @@ async def update_assignment_due_date(self, assignment: AssignmentModel, due_date
...
         return assignment
+    
+    async def delete_assignment(self, assignment: AssignmentModel):
+        from app.services import GiteaService, CourseService, FileOperation, FileOperationType
+        gitea_service = GiteaService()
+        course_service = CourseService(self.session)
+
+        master_repository_name = await course_service.get_master_repository_name()
+        owner = await course_service.get_instructor_gitea_organization_name()
+        assignment_folder = f"{ assignment.name }_{ assignment.id }-dist"
```
frostyfan109: What is this path? It should just delete the `directory_path` of the assignment in the repository.

app/services/canvas_service.py
```diff
@@ -0,0 +1,161 @@
...
+            }
+
+            response = self.session.put(url, headers=headers, json=payload)
+            response.raise_for_status()
+            return response.json()
+
+        except Exception as e:
+            raise HTTPException(status_code=500, detail=str(e))
+        
+    async def update_assignment(self, assignment_id: int, payload: dict):
```
frostyfan109: what is payload??

app/services/canvas_service.py
```diff
@@ -0,0 +1,161 @@
...
+            response = self.session.get(url)
+            response.raise_for_status()
+            return response.json()
+        
+        except requests.RequestException as e:
+            raise LMSNoStudentsFetchedException()
+        except Exception as e:
+            raise HTTPException(status_code=500, detail=str(e))
+
+    async def upload_grades(self, assignment_id: int, user_id: int, grade: float):
```
frostyfan109: calling this upload_grades is kinda misleading considering it is uploading a single grade

app/services/course_service.py
```diff
@@ -60,4 +64,13 @@ async def get_instructor_gitea_organization_name(self) -> str:
     
     async def get_master_repository_name(self) -> str:
         course = await self.get_course()
-        return f"{ course.name }-class-master-repo"
\ No newline at end of file
+        return f"{ course.name }-class-master-repo"
+    
+    async def update_course(self, name: str, start_at: datetime, end_at: datetime) -> CourseModel:
```
frostyfan109: start_at, end_at is nullable elsewhere. Also, update_course should be broken into update_course_name, update_course_start_at, update_course_end_at

app/services/lms_sync_service.py
```diff
@@ -0,0 +1,241 @@
+from datetime import datetime, timedelta
+import json
+from fastapi import Depends, HTTPException
```
frostyfan109, line 3: Don't raise HTTPException. Create some generic 500 canvas service error or something instead. You can instantiate exceptions that inherit CustomException with your own message as the first argument.

app/services/lms_sync_service.py
```diff
@@ -0,0 +1,241 @@
...
+    async def downsync(self):
+        await asyncio.gather(
+            self.sync_course(),
+            self.sync_assignments(),
+            self.sync_students(),
+            self.sync_instructors()
+        )
+
+        return {
+            "message": "Successfully synced the LMS with the database"
```
frostyfan109, line 197: Weird format to return, I'd prefer it just to be void

app/services/lms_sync_service.py
```diff
@@ -0,0 +1,241 @@
...
+                await self.canvas_service.associate_pid_to_user(user_info.onyen, pid)
+
+        return canvas_instructors
+
+    async def upload_grades_from_csv(self, grade_csv):
+        df = pd.read_csv(io.StringIO(grade_csv.decode("utf-8")))
+        
+        for index, row in df.iterrows():
+            try:
+                assignment_id = self._extract_id(row['file'])
```
frostyfan109: This is bad. `assignment_id` should 100% be an argument of this method.

app/services/user/user_service.py
```diff
@@ -1,5 +1,5 @@
 from sqlalchemy.orm import Session
-from app.models import UserModel, AutoPasswordAuthModel
+from app.models import UserModel, AutoPasswordAuthModel, OnyenPIDModel
```
frostyfan109: Onyen and pid should be unassociated from user during deletion.

app/api/api_v1/endpoints/lms_router.py
```diff
@@ -0,0 +1,45 @@
...
+    perm: None = Depends(PermissionDependency(UserIsInstructorPermission))
+):
+    return await LmsSyncService(db).sync_assignments()
+
+@router.post("/lms/grades")
+async def post_grades(
+    *,
+    db: Session = Depends(get_db),
+    perm: None = Depends(PermissionDependency(UserIsInstructorPermission)),
+    file: UploadFile = File(...)
```
frostyfan109, line 43: Honestly it may be easier here to just treat the grades CSV as a string parameter of the JSON body, I don't particularly care though. It should also take an assignment id parameter.

app/services/canvas_service.py
ptlharit2: the payload is the assignment PUT request body - I can change payload to `assignment_body` which might be more descriptive

app/services/user/user_service.py
ptlharit2: I am doing this within the student/instructor sync methods already. 

It belongs more in LMS Sync - the whole reason we are associating/associating pids is because Canvas returns a pid. It's LMS specific logic

app/api/api_v1/endpoints/lms_router.py
ptlharit2: Done - added assignment id as param
Though I'm leaving the csv as an uploadable file - we can revert if this turns out to be a hassle

app/services/lms_sync_service.py
ptlharit2: done

app/services/lms_sync_service.py
ptlharit2: done

app/services/lms_sync_service.py
ptlharit2: done

app/services/course_service.py
ptlharit2: done

app/services/canvas_service.py
ptlharit2: done

app/services/canvas_service.py
ptlharit2: done

app/services/assignment_service.py
ptlharit2: done

app/services/assignment_service.py
ptlharit2: done

app/services/assignment_service.py
ptlharit2: done

app/services/assignment_service.py
ptlharit2: done

app/api/api_v1/endpoints/assignment_router.py
ptlharit2: Oh I wasn't aware JLP had the functionality to change names as well. I changed the service method as well to allow it

app/api/api_v1/endpoints/assignment_router.py
ptlharit2: done

app/api/api_v1/endpoints/assignment_router.py
frostyfan109: It doesn't have that functionality. But the endpoint allows it so...

app/services/user/user_service.py
frostyfan109: That's a valid point, I would prefer it in LMS sync. The worry here is that user deletion outside of LMS sync will no longer work, since the onyen pid model has a foreign key to the user's onyen. Thus, it will error if you try to delete the user before deleting the onyen-pid association (i.e. through the user service, rather than the LMS service).

frostyfan109: For now, you should add it to user deletion as well (duplicate the changes). Make sure it's in a try/catch since it will throw if the associated was already deleted (as it will be). Once implemented, this will be superseded by the cascade delete ticket, which will handle all of this automatically automatically, so we can delete it from both places.

app/services/lms_sync_service.py
ptlharit2: added a new migration to make the dates nullable and changed the service method, you had a good point

ptlharit2:
This PR was merged into another branch to resolve merge conflicts, which was then merged into develop - https://github.com/helxplatform/eduhelx-grader-api/pull/37

All changes from this branch have been merged into develop, hence I am closing this PR.

PR #35, opened 2024-05-06
JLP/JLS Upgrades

frostyfan109:
Upgrades required for JLP/JLS upgrades (asides from a few features that will come in the future).
- `IMPERSONATE_USER` option (only in DEV phase) allows you to launch the Grader API as a fixed user to circumvent auth.
- Update README
- Add `fork_cloned` option so we know if the student has moved their cloned fork somewhere on their filesystem (safeguard so we don't redownload it and confuse the student, this may be replaced down the line)
- Add `fork_remote_url` option so that JLS can clone the student fork for the student automatically.
- Fix permission on list own assignments endpoint, fix permission on get own user endpoint
- Restrict password charset to base64, instead of alphanum + punctuation. Our passwords are 64 characters so still secure.
- Add submission download endpoint (archive of the repository at a treeish id, optionally only files under a path of the repo)
- Add InstructorAssignment, add InstructorAssignmentService for JLP/change list assignments endpoint to work for both students and profs
- Add `add_collaborator`, `remove_collaborator`, `delete_user`, `modify_repo`, `download_repo`, `modify_repository_files` methods to gitea service
- Add `delete credential secret` to K8s service
- Change master repo to be private, add students as readonly to master repo, so student forks are private
- Add `delete_user` method to UserService which cleans up DB, k8s secrets, gitea stuff
- Add appstore-based authentication endpoint
- Add SSH key registration endpoint 
- Add settings endpoint which includes Gitea SSH URL for overriding default ingress url
- Make student forks private by default so that students cannot view others' repositories
- Change CORS policy to only be permissive in dev mode


app/core/utils/header.py
```diff
@@ -0,0 +1,8 @@
+from typing import Tuple
+from email.message import EmailMessage
+
+""" https://docs.python.org/3.11/library/cgi.html#cgi.parse_header """
```
ptlharit2, line 4: do we need this?

app/services/user/student_service.py
```diff
@@ -47,16 +47,60 @@ async def create_student(
...
+            new_name=student_repo_name
+        )
+        self.session.commit()
+
+    async def delete_user(
+        self,
+        onyen: str
+    ) -> None:
+        from app.services import SubmissionService, AssignmentService
+        from app.models import ExtraTimeModel
```
ptlharit2, line 78: I copied over the delete user in student service & user service into my PR. Just to keep in mind when we merge them both

ptlharit2 (approved):

PR #38, opened 2024-06-04
Fix up grading stuff, couple changes to canvas

frostyfan109:


app/services/lms_sync_service.py
```diff
@@ -148,37 +146,28 @@ async def sync_instructors(self):
...
+    async def upload_grades(self, assignment_id: int, grades: list[dict]):
+        for row in grades:
+            user_pid = await self.canvas_service.get_pid_from_onyen(row['onyen'])
+            students = await self.canvas_service.get_users({
+                "enrollment_type": "student"
+            })
+            
+            student = next((student for student in students if student['sis_user_id'] == user_pid), None)
+            if student is None:
+                raise LMSUserNotFoundException()
```
ptlharit2, line 158: We probably want to validate an assignment exists in Canvas similar to how we are validating a student exists in Canvas

ptlharit2 (approved):

app/services/lms_sync_service.py
frostyfan109: Yeah that can wait. But let me note here that the canvas service itself should be doing this, not the LMS service.

ptlharit2: Yes validation should be moved to the canvas service. Sure lets handle it after

frostyfan109: While we're at it, let me mention here that the Canvas service should not be a pure API wrapper that only allows you to invoke endpoints as they are exposed by canvas. That's fine to have in the service (should probably be private, i.e. prefixed with _) but other services for example should not be invoking get_users with enrollment_type student, there should just be a get_students method. Generally, none of the public methods on the Canvas service should require innate knowledge of the API endpoint, those should be handled by the arguments.

PR #42, opened 2024-06-11
Merge control

frostyfan109:
- Adds merge control Git hook
- Adds CRUD events
- Refactors update_assignment and update_course methods
- Adds CleanupService which rectifies state (e.g. with Gitea, Kubernetes) if creation/deletion of course/users fails partway through
- Adds ssh folder to gitignore
- Validate due date is not before available_date, when applicable


app/api/api_v1/endpoints/lms_router.py
```diff
@@ -23,7 +23,7 @@ async def downsync(
     db: Session = Depends(get_db),
     perm: None = Depends(PermissionDependency(UserIsInstructorPermission))
 ):
-    return await LmsSyncService(db).downsync()
```
Hoid, line 26: Why remove the return here?

app/main.py
```diff
@@ -22,10 +22,19 @@ def init_routers(app: FastAPI):
...
+        content = { "error_code": exc.error_code, "message": exc.message }
+        if settings.DEV_PHASE == DevPhase.DEV:
+            content["stack"] = exc.stack
         return JSONResponse(
             status_code=exc.code,
-            content={"error_code": exc.error_code, "message": exc.message},
+            content=content,
         )
+    
+def init_monkeypatch():
```
Hoid, line 33: The `monkeypatch` part of the method name is fine, but I think the `init` part should be more descriptive. What is this `init`ing? I don't quiiite understand the comment either. What changes are you making to the serializers for custom types, exactly?

app/models/assignment.py
```diff
@@ -9,6 +9,7 @@
 from sqlalchemy.ext.hybrid import hybrid_property
 from app.database import Base
 from .user import StudentModel
+from app.core.exceptions import AssignmentDueBeforeOpenException
```
Hoid: I don't think this is used

app/models/user/onyen_pid.py
```diff
@@ -1,8 +1,8 @@
 from sqlalchemy import Column, String, Text, ForeignKey
+from sqlalchemy.orm import relationship
```
Hoid, line 2: I don't think this is used

app/events/handlers.py
```diff
@@ -14,12 +14,36 @@
...
+    student_service = StudentService(session)
+    course_service = CourseService(session)
+    gitea_service = GiteaService(session)
+
+    course = await course_service.get_course()
+    students = await student_service.list_students()
+    hook_content = await gitea_service.get_merge_control_hook()
+    for student in students:
+        student_repo_name = course_service._compute_student_repository_name(course.name)
+        await GiteaService(session).set_git_hook(
```
Hoid, line 33: do you not want to reuse the `gitea_service` that you created on line 26 here, and below on line 42?

app/services/assignment_service.py
```diff
@@ -118,29 +131,76 @@ async def get_assignment_by_name(self, name: str) -> AssignmentModel:
             raise AssignmentNotFoundException()
         return assignment
     
-    async def update_assignment_name(self, assignment: AssignmentModel, new_name: str) -> AssignmentModel:
-        assignment.name = new_name
-        assignment.last_modified_date = func.current_timestamp()
+    async def update_assignment(self, assignment: AssignmentModel, update_assignment: UpdateAssignmentSchema) -> AssignmentModel:
+        update_fields = update_assignment.dict(exclude_unset=True)
+        
+        if "name" in update_fields:
```
Hoid, line 137: I hate having to go through every field in this but I can't think of a better way to do it right now

app/services/assignment_service.py
```diff
@@ -118,29 +131,76 @@ async def get_assignment_by_name(self, name: str) -> AssignmentModel:
...
-        assignment.due_date = due_date
-        assignment.last_modified_date = func.current_timestamp()
-        self.session.commit()
-        return assignment
+    async def get_master_notebook_name(self, assignment: AssignmentModel) -> str:
+        return self._compute_master_notebook_name(assignment.name)
+    
+    @staticmethod
+    def _compute_master_notebook_name(self, assignment_name: str) -> str:
+        return f"{ assignment_name }-prof.ipynb"
```
Hoid, line 183: Why the `-prof`? Won't the master notebook name be copied exactly, with its filename, to all of the forks? Or will the filename change?

app/services/cleanup_service.py
```diff
@@ -0,0 +1,54 @@
+from sqlalchemy.orm import Session
+from app.models import UserModel, CourseModel
+from app.services import GiteaService, KubernetesService
+
+class CleanupService:
```
Hoid, line 5: Not sure I quite like the idea of this CleanupService. You're just abstracting away just one or two potential lines of code from an original method that would instead have to clean up itself, not several lines or multiple methods. Plus you only implemented this for the Course and User cleanup, not the Assignment cleanup for example [here](https://github.com/helxplatform/eduhelx-grader-api/pull/42/files#diff-7d9bf52ff181cb543f98215404d4a2365134ecbca373f0076f483724efcb3affR78). Is there a reason this CleanupService is needed instead of always doing it in an `except` clause, like you did in the Assignment cleanup?

app/api/api_v1/endpoints/lms_router.py
frostyfan109: Because the downsync method doesn't return anything either

app/main.py
frostyfan109: It's initializing monkey patches?

frostyfan109: We're monkey patching the pydantic JSON module's serializers ("encoder") to serialize a custom type (unset)

app/events/handlers.py
frostyfan109: probably do, it's inconsequential though since services are built to be stateless

app/services/assignment_service.py
frostyfan109: You could probably write a utility function to do this given the schema class (as in UpdateAssignmentSchema here). I'm sure if you look at pydantic documentation (or dir some objects) there's a way to list the keys of a schema

frostyfan109: Or just, for key in update_fields: getattr(model, key) = update_fields[key] and then override it in cases where the key is for some reason not the same as the database model key

app/services/assignment_service.py
frostyfan109: I've actually changed it to just {name}.ipynb in my new branch, and then the student notebook is {profnotebookname}-student.ipynb.

Also, the master notebook is private and not added to the repo/visible to students. This is because the master notebook contains the test cases

app/services/cleanup_service.py
frostyfan109: Because cleanup has to be done in stages based on how far along the service method got in its execution before encountering an unrecoverable failure. The create assignment method only has 1 stage of cleanup (database deletion) and so does not need a whole method dedicated to cleanup.

Hoid (approved):

PR #48, opened 2024-07-17
Grading

frostyfan109:
- Add master_notebook_path, student_notebook_path for grading
- Add grading functionality using non-containerized otter_run locally (GradingService)
- Define "active" submission as latest and implement various active submission functionalities
- Add requirements.txt and default notebook with otter config to assignments during creation
- Disable deleting files in merge control policy
- Add protected files policy pre-receive hook
  - Add hook combiner (run multiple scripts for a single type of hook) so that we can run protected files and merge control policies as pre-receives


app/api/api_v1/endpoints/assignment_router.py
```diff
@@ -68,4 +73,24 @@ async def get_assignments(
...
+    grading_body: GradingBody,
+    perm: None = Depends(PermissionDependency(UserIsInstructorPermission))
+):
+    assignment = await AssignmentService(db).get_assignment_by_name(assignment_name)
+    grade_report = await GradingService(db).grade_assignment(
+        assignment,
+        grading_body.master_notebook_content,
+        grading_body.otter_config_content
+    )
+    print(grade_report)
```
Hoid: Do you want this `print` statement in there?

app/api/api_v1/endpoints/lms_router.py
```diff
@@ -21,7 +21,7 @@ class UploadGradesBody(BaseModel):
 async def downsync(
     *,
     db: Session = Depends(get_db),
-    perm: None = Depends(PermissionDependency(UserIsInstructorPermission))
+    # perm: None = Depends(PermissionDependency(UserIsInstructorPermission))
```
Hoid: Should remove unused commented out code

app/services/assignment_service.py
```diff
@@ -174,33 +199,36 @@ async def get_latest_due_date(self, assignment: AssignmentModel) -> datetime | N
...
+
+### Protected ###
+{ protected_files }
 """
+    
+    """
+    NOTE: File paths are not necessarily real files and may instead be globs.
+    NOTE: File paths are relative to `assignment.directory_path`.
+    """
+    async def get_protected_files(self, assignment: AssignmentModel) -> str:
```
Hoid, line 229: This doesn't need to be async 😂

app/schemas/submission.py
```diff
@@ -3,12 +3,16 @@
 from .user import StudentSchema
 from .assignment import AssignmentSchema
 
-class SubmissionSchema(BaseModel):
+class DatabaseSubmissionSchema(BaseModel):
```
Hoid, line 6: Why the separation between `DatabaseSubmissionSchema` and `SubmissionSchema` here? Why not just put the `active` flag on the original `SubmissionSchema` object?

app/schemas/grade_report.py
```diff
@@ -0,0 +1,32 @@
...
+    id: int
+    average: float
+    median: float
+    minimum: float
+    maximum: float
+    stdev: float
+    scores: list[float]
+    total_points: float
+    num_submitted: int
+    # num_skipped = num_submitted (== num students w/ submissions) - num_active_submissions_already_graded
```
Hoid, line 15: Any reason for this to be here? If so I think it should have a comment explaining why it's commented out

app/schemas/submission.py
```diff
@@ -3,12 +3,16 @@
...
     id: int
     commit_id: str
+    graded: bool
     submission_time: datetime
     # student: StudentSchema
     # assignment: AssignmentSchema
```
Hoid: Remove commented code

app/api/api_v1/endpoints/lms_router.py
frostyfan109: good catch, this isn't supposed to be commented.

app/services/assignment_service.py
frostyfan109: Yeah but there is a world in which this function needs to be async down the line

app/schemas/submission.py
frostyfan109: It's the same concept as StudentAssignmentSchema and InstructorAssignmentSchema. I would eventually like to go through and refactor all the base schemas into `Database{SchemaName}` so that it's more clear.

Basically, active has to be computed using a query, so it has to be added in when creating SubmissionSchema, whereas DatabaseSubmissionSchema can be created directly from a SubmissionModel.

app/schemas/grade_report.py
frostyfan109: it's not commented out, it is a comment

Hoid: Oh gotcha, valid

app/events/dispatcher.py
```diff
@@ -0,0 +1,10 @@
+import logging
+from fastapi_events.dispatcher import dispatch as _dispatch
+
+logger = logging.getLogger(__file__)
+
+def dispatch(*args, **kwargs):
```
Hoid: I think this is already in develop? Maybe try doing `git pull origin develop` on your branch, I think some of these file changes will go away

app/api/api_v1/endpoints/submission_router.py
```diff
@@ -59,38 +69,33 @@ async def get_own_submissions(
...
     assignment = await AssignmentService(db).get_assignment_by_id(assignment_id)
-    submissions = await SubmissionService(db).get_submissions(student, assignment)
+    submissions = await submission_service.get_submissions(student, assignment)
 
-    return submissions
+    return [await submission_service.get_submission_schema(s) for s in submissions]
 
-@router.get("/latest_submission/", response_model=SubmissionSchema)
-async def get_latest_submission(
+@router.get("/active_submission/", response_model=SubmissionSchema)
```
Hoid: These endpoint URIs seem like they need to be cleaned up. I feel like this one should be `GET /submissions/active` and the one below on line 126 should be `GET /submissions/active/download` (well, I think they should be completely different and use path params instead of query params, but this would clean them up with minimal changes).

frostyfan109: fixing this

app/services/course_service.py
```diff
@@ -39,10 +39,11 @@ async def create_course(self, name: str, start_at: datetime = None, end_at: date
             raise CourseAlreadyExistsException()
         except NoCourseExistsException:
             pass
-
+        
+        print("CREATING COURSE")
```
Hoid, line 43: Did you mean to keep all the print statements in this file? If so shouldn't it be done by the logger instead of print()?

frostyfan109: I'll change these to logger debugs. I think they're useful to leave in the code for now at least in case any bugs with syncing crop up again.

frostyfan109: On second thought, I had asked Harit to work on getting the logger functional outside of app context. I'm not sure what he's done with that, or if we're just supposed to use module-level loggers or what. These should stay prints until then, since downsync runs outside of the app context (i.e. we downsync right before we start the uvicorn server), so trying to access the app logger would cause errors.

app/events/dispatcher.py
frostyfan109: I'll do a merge

app/services/canvas_service.py
```diff
@@ -82,6 +98,58 @@ async def get_assignments(self):
...
+        student_id: int | None = None,
+        **kwargs
+    ):
+        student_ids = [student_id] if student_id is not None else []
+        submissions = await self.get_submissions_for_assignments([assignment_id], student_ids, **kwargs)
+        if student_id is not None: return submissions[0]
+        return submissions
+    
+    """ Get the current attempt count of the student for the assignment. If the student hasn't submitted, then 0. """
+    async def get_current_submission_attempt(
```
Hoid: This would be a lot clearer I think if it was plural (get_current_submission_attempts). I had to read the method to realize what it meant lol

frostyfan109: fixed

Hoid (approved):
There's a couple small parts of the grading code that I just don't understand how they work (we can talk offline if I need to understand it later), but as long as you're confident in it I'm good to merge this.

PR #50, opened 2024-07-19
Overhaul session transactions, cleanup event dispatches, replace fastapi-events

frostyfan109:
- Replace fastapi-events with pymitter. fastapi-events couldn't be used outside the app context which was problematic
- Overhaul our database writes with commit transactions for improved cleanup
- Try/catch event dispatches, which are subject to failure (they were with fastapi-events too, we just weren't catching them)


Hoid (approved):

frostyfan109:
Websocket-pubsub supercedes

PR #56, opened 2024-08-02
Add logic to support workers in startup

Hoid:
Configure startup logic so that the number of workers to be used can be passed in correctly.

Credit to Paras for a lot of this work!

### DIFF
--- a/start.py
+++ b/start.py
@@ -71,7 +71,7 @@ def main(host, port, reload, workers):
         if workersInt is not None and workersInt <= 0: raise ValueError()
     except ValueError:
         raise SystemExit("Error: Please use a positive integer value for the number of workers.")
-    
+
     # Start the application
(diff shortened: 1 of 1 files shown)


start.py
```diff
@@ -53,22 +52,43 @@ def main(host, port, reload):
...
+    try:
+        if port: portInt = int(port)
+        if portInt is not None and portInt <= 0: raise ValueError()
+    except ValueError:
+        raise SystemExit("Error: Please use a positive integer value for the port.")
+    try:
+        if workers: workersInt = int(workers)
+        if workersInt is not None and workersInt <= 0: raise ValueError()
+    except ValueError:
+        raise SystemExit("Error: Please use a positive integer value for the number of workers.")
```
ptlharit2, line 73: minor improvement - we could reduce duplicate code here

ptlharit2 (approved):

PR #57, opened 2024-08-07
Use Canvas's status of "published" internally 

Hoid:


ptlharit2 (approved):

ptlharit2:
looks great, and glad you also changed the unit tests!

app/services/assignment_service.py
```diff
@@ -177,6 +182,9 @@ async def update_assignment(self, assignment: AssignmentModel, update_assignment
         if "due_date" in update_fields:
             assignment.due_date = update_fields["due_date"]
 
+        if "is_published" in update_fields:
+            assignment.is_published = update_fields["is_published"]
```
frostyfan109, line 186: you need to add this to both UpdateAssignmentSchema and the schema used for the API endpoint handler that updates assignments.

app/services/assignment_service.py
```diff
@@ -249,22 +257,22 @@ def __init__(self, session: Session, instructor: InstructorModel, assignment: As
...
         current_timestamp = self.session.scalar(func.current_timestamp())
         return current_timestamp >= self.assignment.available_date
     
-    def get_is_closed(self) -> bool:
-        if not self.assignment.is_created: return False
+    def _get_is_closed(self) -> bool:
+        if not self.assignment.is_published: return False
 
         current_timestamp = self.session.scalar(func.current_timestamp())
         return current_timestamp > self.assignment.due_date
```
frostyfan109, line 270: here and also in get_is_available I don't think this is the right methodology. Canvas considers an assignment always available to work on if it is both published and is missing either an open or due date. Instead, this will likely just error if the open/due date are unset, since you can't compare a NoneType to an int. The is_created hybrid_property was performing this check for our previously.

app/services/assignment_service.py
```diff
@@ -307,34 +315,30 @@ def get_adjusted_due_date(self) -> datetime | None:
...
+    def _get_is_available(self) -> bool:
         current_timestamp = self.session.scalar(func.current_timestamp())
         return current_timestamp >= self.get_adjusted_available_date()
     
-    def get_is_closed(self) -> bool:
-        if not self.assignment.is_created: return False
-
+    def _get_is_closed(self) -> bool:
         current_timestamp = self.session.scalar(func.current_timestamp())
         return current_timestamp > self.get_adjusted_due_date()
```
frostyfan109, line 324: same issue here as in instructorassignmentservice's methods. comparing a nullish to an int

frostyfan109 (changes requested):

app/services/assignment_service.py
Hoid: Yep good catch

PR #64, opened 2024-08-14
Create assignment status enum to synchronize is_published usage with Canvas

Hoid:


app/core/config.py
```diff
@@ -70,7 +70,7 @@ class Settings(BaseSettings):
     SQLALCHEMY_DATABASE_URI: Optional[PostgresDsn] = None
 
 
-    @validator("IMPERSONATE_USER", pre=True)
+    @validator("IMPERSONATE_USER", pre=True, allow_reuse=True)
```
frostyfan109: remove

app/services/assignment_service.py
```diff
@@ -74,11 +85,14 @@ async def create_assignment(
...
+        
+        # See comment above commented out FileOperation below
+        # master_notebook_content = json.dumps({ "cells": [otter_config_cell, title_cell], "metadata": {  "kernelspec": {   "display_name": "Python 3 (ipykernel)",   "language": "python",   "name": "python3"  },  "language_info": {   "codemirror_mode": {    "name": "ipython",    "version": 3   },   "file_extension": ".py",   "mimetype": "text/x-python",   "name": "python",   "nbconvert_exporter": "python",   "pygments_lexer": "ipython3",   "version": "3.11.5"  } }, "nbformat": 4, "nbformat_minor": 5})
 
         gitignore_path = f"{ directory_path }/.gitignore"
         gitignore_content = await self.get_gitignore_content(assignment)
         
+        # See comment above commented out FileOperation below
```
frostyfan109: ?

app/services/assignment_service.py
```diff
@@ -185,6 +199,14 @@ async def update_assignment(self, assignment: AssignmentModel, update_assignment
         if "due_date" in update_fields:
             assignment.due_date = update_fields["due_date"]
 
+        if "is_published" in update_fields:
+            if update_fields["is_published"] == False:
+                canvas_service = CanvasService(self.session)
+                submissions = await canvas_service.get_submissions(assignment.id)
+                if len([sub for sub in submissions if sub["workflow_state"] != "unsubmitted"]) > 0:
+                    raise AssignmentAlreadyHasSubmissions("assignment")
```
frostyfan109: "assignment" is not a descriptive error message

app/services/assignment_service.py
```diff
@@ -267,60 +289,88 @@ def _compute_master_notebook_name(assignment_name: str) -> str:
...
-        assignment = AssignmentSchema.from_orm(self.assignment).dict()
-        assignment["protected_files"] = await self.get_protected_files(self.assignment)
-        assignment["overwritable_files"] = await self.get_overwritable_files(self.assignment)
+        assignment = AssignmentSchema.from_orm(self.assignment_model).dict()
+        course = CourseSchema.from_orm(self.course_model).dict()
+        current_timestamp = self.session.scalar(func.current_timestamp())
 
-        assignment["is_available"] = self.get_is_available()
-        assignment["is_closed"] = self.get_is_closed()
+        assignment_status = AssignmentStatus.get_value_for(assignment, course, current_timestamp)
```
frostyfan109: you aren't setting the `status` field

app/services/assignment_service.py
```diff
@@ -267,60 +289,88 @@ def _compute_master_notebook_name(assignment_name: str) -> str:
...
             ) \
             .first()
         return extra_time_model
 
     # The release date for a specific student, considering extra_time
     def get_adjusted_available_date(self) -> datetime | None:
-        if self.assignment.available_date is None: return None
-
-        deferred_time = self.extra_time_model.deferred_time if self.extra_time_model is not None else timedelta(0)
+        assignmentOpenDate = self.assignment_model.available_date or self.course_model.start_at
```
frostyfan109: camelCase?

app/services/assignment_service.py
```diff
@@ -267,60 +289,88 @@ def _compute_master_notebook_name(assignment_name: str) -> str:
...
+        return assignmentOpenDate + deferred_time
 
-        current_timestamp = self.session.scalar(func.current_timestamp())
-        return current_timestamp >= self.assignment.available_date
-    
-    def get_is_closed(self) -> bool:
-        if not self.assignment.is_created: return False
+    # The due date for a specific student, considering extra_time
+    def get_adjusted_due_date(self) -> datetime | None:
+        assignmentDueDate = self.assignment_model.due_date or self.course_model.end_at
```
frostyfan109: camelCase

app/services/assignment_service.py
```diff
@@ -267,60 +289,88 @@ def _compute_master_notebook_name(assignment_name: str) -> str:
...
-        self.assignment = assignment
+        self.instructor_model = instructor_model
+        self.assignment_model = assignment_model
+        self.course_model = course_model 
 
-    def get_is_available(self) -> bool:
-        if not self.assignment.is_created: return False
+    # The release date for a specific student, considering extra_time
+    def get_adjusted_available_date(self) -> datetime | None:
+        assignmentOpenDate = self.assignment_model.available_date or self.course_model.start_at
```
frostyfan109: camelCase

app/services/assignment_service.py
frostyfan109: get_value_for?

app/services/assignment_service.py
```diff
@@ -267,60 +289,88 @@ def _compute_master_notebook_name(assignment_name: str) -> str:
...
+        if assignmentOpenDate is None: return None
+        deferred_time = self.extra_time_model.deferred_time or timedelta(0)
         
-        return self.assignment.available_date + deferred_time
+        return assignmentOpenDate + deferred_time
 
     # The due date for a specific student, considering extra_time
     def get_adjusted_due_date(self) -> datetime | None:
-        if self.assignment.due_date is None: return None
+        assignmentDueDate = self.assignment_model.due_date or self.course_model.end_at
```
frostyfan109: camelCase

app/services/assignment_service.py
```diff
@@ -331,51 +381,65 @@ def get_adjusted_due_date(self) -> datetime | None:
...
 
-        if not self.get_is_available():
+        if assignment_status == AssignmentStatus.UPCOMING:
             raise AssignmentNotOpenException()
 
-        if self.get_is_closed():
+        if assignment_status == AssignmentStatus.CLOSED:
             raise AssignmentClosedException()
         
+        from app.services.submission_service import SubmissionService
```
frostyfan109: import in the middle of a function is bad

app/services/assignment_service.py
```diff
@@ -331,51 +381,65 @@ def get_adjusted_due_date(self) -> datetime | None:
...
-            self.student,
-            self.assignment
+            self.student_model,
+            self.assignment_model
         )
         assignment["adjusted_available_date"] = self.get_adjusted_available_date()
         assignment["adjusted_due_date"] = self.get_adjusted_due_date()
-        assignment["is_available"] = self.get_is_available()
-        assignment["is_closed"] = self.get_is_closed()
+        assignment["is_available"] = assignment_status == AssignmentStatus.OPEN
```
frostyfan109, line 444: youre not setting `status` field

app/api/api_v1/endpoints/assignment_router.py
```diff
@@ -24,6 +25,7 @@ class UpdateAssignmentBody(BaseModel):
     max_attempts: PositiveInt | None
     available_date: datetime | None
     due_date: datetime | None
+    is_published: bool = UNSET
```
frostyfan109, line 28: this isn't used

Hoid: image

app/core/config.py
Hoid: Fixed

app/services/assignment_service.py
Hoid: Fixed

app/services/assignment_service.py
Hoid: Fixed

Hoid: Fixed

app/services/assignment_service.py
Hoid: I think this is fixed

app/services/assignment_service.py
Hoid: Fixed

app/services/assignment_service.py
```diff
@@ -267,60 +290,77 @@ def _compute_master_notebook_name(assignment_name: str) -> str:
...
+
+    def get_assignment_status(self) -> AssignmentStatus:
+        if not self.assignment_model.is_published: return AssignmentStatus.UNPUBLISHED
 
         current_timestamp = self.session.scalar(func.current_timestamp())
-        return current_timestamp > self.assignment.due_date
+        adjusted_available_date = self.get_adjusted_available_date()
+        adjusted_due_date = self.get_adjusted_due_date()
+        
+        if adjusted_available_date >= current_timestamp: return AssignmentStatus.UPCOMING
```
ptlharit2, line 324: this comparison will cause an error if `adjusted_available_date` is None. which is one of the possibilities of values that can be returned by `get_adjusted_available_date`

app/services/assignment_service.py
```diff
@@ -267,60 +290,77 @@ def _compute_master_notebook_name(assignment_name: str) -> str:
...
+        if adjusted_due_date > current_timestamp: return AssignmentStatus.OPEN
```
ptlharit2, line 326: same here

app/services/assignment_service.py
```diff
@@ -185,8 +201,15 @@ async def update_assignment(self, assignment: AssignmentModel, update_assignment
         if "due_date" in update_fields:
             assignment.due_date = update_fields["due_date"]
 
+        if "is_published" in update_fields:
+            if update_fields["is_published"] == False:
+                canvas_assignment = await LmsSyncService(self.session).get_assignment(assignment.id)
+                if canvas_assignment["unpublishable"] == False:
+                    raise AssignmentCannotBeUnpublished
+            assignment.is_published = update_fields["is_published"]
```
ptlharit2, line 208: In the professor extension, @frostyfan109 when a open/due date is changed, and we want to sync it up to Canvas - we will also have to update our local instance assignment status.

@Hoid you will need to update the `is_published` status for an assignment at line 201 as well. And also at line 198 when `available_date` is changed

app/services/assignment_service.py
```diff
@@ -331,51 +371,64 @@ def get_adjusted_due_date(self) -> datetime | None:
...
-
-    def get_is_available(self) -> bool:
-        if not self.assignment.is_created: return False
+        return assignment_due_date + deferred_time + extra_time + self.student_model.base_extra_time
 
+    def _get_is_available(self) -> bool:
+        adjusted_available_date = self.get_adjusted_available_date()
+        if adjusted_available_date is None: return True
+        current_timestamp = self.session.scalar(func.current_timestamp())
+        return current_timestamp >= adjusted_available_date
```
ptlharit2, line 389: this line will error if `self.get_adjusted_available_date()` returns None. which is possible in the current implementation

app/services/assignment_service.py
```diff
@@ -331,51 +371,64 @@ def get_adjusted_due_date(self) -> datetime | None:
...
+    
+    def _get_is_closed(self) -> bool:
+        adjusted_due_date = self.get_adjusted_due_date()
+        if adjusted_due_date is None: 
+            return not self._get_is_available()
         current_timestamp = self.session.scalar(func.current_timestamp())
-        return current_timestamp >= self.get_adjusted_available_date()
+        return current_timestamp > adjusted_due_date
```
ptlharit2, line 396: this line can error

app/services/assignment_service.py
```diff
@@ -331,51 +371,64 @@ def get_adjusted_due_date(self) -> datetime | None:
...
+    def get_assignment_status(self) -> AssignmentStatus:
+        if not self.assignment_model.is_published: return AssignmentStatus.UNPUBLISHED
 
         current_timestamp = self.session.scalar(func.current_timestamp())
-        return current_timestamp > self.get_adjusted_due_date()
+        adjusted_available_date = self.get_adjusted_available_date()
+        adjusted_due_date = self.get_adjusted_due_date()
+        
+        if adjusted_available_date >= current_timestamp: return AssignmentStatus.UPCOMING
+        if adjusted_due_date > current_timestamp: return AssignmentStatus.OPEN
```
ptlharit2, line 412: above two lines can error

app/services/canvas_service.py
```diff
@@ -7,12 +7,18 @@
...
 from fastapi import HTTPException
 from sqlalchemy.orm import Session
 from app.core.config import settings
+from app.core.exceptions.assignment import AssignmentNotPublishedException
+from app.enums.assignment_status import AssignmentStatus
+from app.enums.canvas.canvas_workflow_state_filter import CanvasWorkflowStateFilter
 from app.models import UserModel, OnyenPIDModel
+from app.models.assignment import AssignmentModel
+from app.models.course import CourseModel
+from app.models.user.student import StudentModel
```
ptlharit2: above imports aren't used anywhere in this file I think

app/services/assignment_service.py
frostyfan109: not sure i'm totally following this. the assignment status is computed, rather than stored, so after sync that status will recompute based on the status of published, available_date, due_date, current_time, etc.

app/services/assignment_service.py
frostyfan109: there's a null check (378)? not sure what you mean

app/services/assignment_service.py
frostyfan109: there's a null check

app/services/assignment_service.py
ptlharit2: literally didnt see that if. ignore me I need my glasses checked

app/services/assignment_service.py
ptlharit2: in the `update_assignment` method, when we update the available/due date of an assignment, we also need to update the is_published field to match do we not.

current implementation is not handling this?

Hoid: @ptlharit2 no, the `is_published` field is strictly a copy of the `published` field that Canvas has on assignments, just in our DB. and like Griffin said the `AssignmentStatus` is always computed, not stored in the database. so we don't ever have to update that value, we only have to update the values it depends on, like `due_date` and `is_published`.

app/services/canvas_service.py
Hoid: Fixed

app/services/assignment_service.py
Hoid: Check this logic now, I think I added a sufficient null check

app/services/assignment_service.py
Hoid: I think I fixed this

app/services/assignment_service.py
Hoid: The decision I made is that if an assignment doesn't have a available date _or_ a course open date set (which is how `adjusted_available_date` becomes None) then I'm putting the assignment in the UPCOMING status, and if it does have an available date of some form but doesn't have a due date _or_ a course end date set (which is how adjusted_due_date becomes None) then I mark the assignment with status CLOSED. That way it never accidentally becomes an assignment the student can work on when we haven't set up the course correctly yet. Fixing this is easy in the configuration: just set a course start and end date at a minimum.

Hoid: @frostyfan109 Tagging you too so you are bought in to this logic

ptlharit2: Yup I like it, not resolving the convo so that we dont lost this, but I am okay with this approach

ptlharit2 (approved):

app/services/assignment_service.py
```diff
@@ -267,60 +289,78 @@ def _compute_master_notebook_name(assignment_name: str) -> str:
...
+        if not self.assignment_model.is_published: return AssignmentStatus.UNPUBLISHED
 
         current_timestamp = self.session.scalar(func.current_timestamp())
-        return current_timestamp > self.assignment.due_date
+        adjusted_available_date = self.get_adjusted_available_date()
+        adjusted_due_date = self.get_adjusted_due_date()
+        
+        if adjusted_available_date is None or adjusted_available_date >= current_timestamp: return AssignmentStatus.UPCOMING
+
+        if adjusted_due_date and adjusted_due_date > current_timestamp: return AssignmentStatus.OPEN
```
frostyfan109: This logic is still incorrect. `IF adjusted_available_date is None`, the assignment has no open date, it is always open. That means it should not be considered upcoming. Perhaps some variables for these would be helpful. Also, comments between lines about the inferred state would be helpful (e.g., after the first line, we know the assignment is definitely published)

`IF adjusted_due_date` <=> `IF adjusted_due_date is not None`, you are checking the assignment has a due date? which is confusing?

frostyfan109: (1) If the assignment isn't published, return UNPUBLISHED. Proceeding, the assignment is definitely published.

(2) If the assignment has no available_date, you can assume it's always available (open date = -Infinity).
- Thus, return UPCOMING if available_date IN FUTURE.
- Note here if available_date is None, the assignment is always available. It can't be UPCOMING.
- Proceeding, the assignment is definitely published, and definitely not UPCOMING (i.e., available_date is definitely in the past)

(3) If the assignment has no due_date, you can assume it's never due (due date = Infinity).
- Thus, return OPEN if due_date IN FUTURE.
- Note here if due_date is None, the assignment is never due. It can't be CLOSED.

What if both available_date and due_date is None? The assignment is always available and never due. It is always OPEN. A proper implementation of the following logic should implicitly lead to this conclusion.

Hoid: The reason I chose to set the assignment to UPCOMING if there is no available_date _or_ course start date set and CLOSED if no due_date _or_ course end date is set is because I don't think it's a good idea for an assignment to accidentally be set to OPEN just because the class isn't configured fully. And there's an easy way to fix each one: set the course start and end dates. That will fix it for all assignments. What you're proposing is to set it to OPEN for both of these scenarios, which I think is a worse option because if the professor hasn't set a course start date and then assignment 2 comes around and they create an assignment without an available date, it will automatically just be set to OPEN right then.

As for the comments within the method, that's a good point, I can do that. And regarding the scenario of if available_date _and_ due_date is None, I just made the decision to handle it sequentially and check first for the available_date, and if that's None (including the course start date) then put the assignment in UPCOMING status, and don't even check the due_date.

frostyfan109: I agree, to some extent. But it's a professor's responsibility to understand how Canvas works. If they create and publish an assignment with no open/due date, then Canvas treats that as an assignment that is always open.

I'm not sure I follow what you're saying initially. Even if the course isn't misconfigured and does have a start/end date set, an assignment without an open/close date would still be automatically open for the entirety of the course. That's effectively the same as just always open anyways.

As an aside: consider a course without start/end dates configured properly, and consider a published assignment with an available date in the past but no due date set. It won't be marked UPCOMING since available date has already past, but due_date is None so it's going to return that the assignment is closed. That doesn't make sense.

If you really want to disallow that sort of behavior, then just always return UPCOMING for everything until both the course start and due date are set. Then implement the ideal logic instead

frostyfan109 (approved):

PR #68, opened 2024-08-16
Fix the pid interaction on the 2U instance in syncing with Canvas

ptlharit2:


app/services/lms_sync_service.py
```diff
@@ -177,6 +191,11 @@ async def upsync_grade(
         comments: str | None = None,
     ):
         user_pid = await self.canvas_service.get_pid_from_onyen(submission.student.onyen)
+        
+        # If this course runs on a 2U Digital Campus instance, append ":UNC" to the PID
+        if "digitalcampus" in settings.CANVAS_API_URL:
```
Hoid, line 196: I think you should put this line before all of the for loops you've added

Hoid: If you disagree though I don't think it's a big deal

frostyfan109:
This is an okay hacky solution for now. In the future, the pid onyen table should store the namespace of the pid mapping. Then the get_pid_from_onyen method could return some sort of object with both pid, namespace, and full sis_user_id (since de-namespaced pid is needed for non-canvas interactions, but namespaced pid (the full `sis_user_id`) is needed for canvas interactions.

ptlharit2:
> This is an okay hacky solution for now. In the future, the pid onyen table should store the namespace of the pid mapping. Then the get_pid_from_onyen method could return some sort of object with both pid, namespace, and full sis_user_id (since de-namespaced pid is needed for non-canvas interactions, but namespaced pid (the full `sis_user_id`) is needed for canvas interactions.

100%. This PR is just a way to get around this inconsistency and deploy to meet the deadline

app/services/lms_sync_service.py
ptlharit2: yeah this check could be added, though the flow would be the same. since it's a temporary addition to this file I am not big on adding more checks, besides we'll remove it in favour of a more robust solution the first chance we get

Hoid (approved):

PR #75, opened 2024-08-26
canvas submission upload at submission time

frostyfan109:
- Upload student submissions to canvas course files at submission time rather than grading time
- Canvas submission upsync at submission time rather than grading time
- Rename submission assignment folders to use assignment name instead of id
- Fix file naming for submission files

### DIFF
--- a/app/services/canvas_service.py
+++ b/app/services/canvas_service.py
@@ -181,29 +181,32 @@ class CanvasService:
         self,
         upload_url: str,
         file: BinaryIO,
-        parent_folder_path_or_id: str | int,
-        on_duplicate: DuplicateFileAction,
-        *,
-        headers={}
+        parent_folder_path_or_id: Path | str | int | None = None,
+        on_duplicate: DuplicateFileAction = DuplicateFileAction.OVERWRITE
     ):
         use_id = isinstance(parent_folder_path_or_id, int)
+        use_path = isinstance(parent_folder_path_or_id, (str, Path))
         
         # Seek to end of file buffer
         file.seek(0, 2)
         payload = {
             "name": file.name,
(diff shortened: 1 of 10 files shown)


app/services/canvas_service.py
```diff
@@ -181,29 +181,32 @@ async def _upload_file(
...
         file: BinaryIO,
-        parent_folder_path_or_id: str | int,
-        on_duplicate: DuplicateFileAction,
-        *,
-        headers={}
+        parent_folder_path_or_id: Path | str | int | None = None,
+        on_duplicate: DuplicateFileAction = DuplicateFileAction.OVERWRITE
     ):
         use_id = isinstance(parent_folder_path_or_id, int)
+        use_path = isinstance(parent_folder_path_or_id, (str, Path))
```
Hoid, line 188: I don't think isinstance does a None check, so you still need to do that since you've added the ability for `parent_folder_path_or_id` to be None.

frostyfan109: if you look a bit further down no action is taken if the value is None.

Hoid: Oh, I was thinking about the use of `use_path` on line 201, but yea if you just do `str(None)` you just get `None` so it's fine

frostyfan109: Maybe I misunderstood what you meant in your first comment. Isinstance definitely checks NoneType. isinstance(None, <anything_other_than_NoneType>) is false. So neihter use_path or use_id will ever be true when the arg is none

Hoid: Oh that's fair too, it does do what I was wanting, yea. Not sure what I thought now 🤷‍♂️

PR #76, opened 2024-08-27
Submission downsync

frostyfan109:
really not too sure about this one...


app/api/api_v1/endpoints/assignment_router.py
```diff
@@ -47,8 +47,6 @@ async def update_assignment_fields(
     update_schema = UpdateAssignmentSchema(**updated_set_fields)
     assignment = await AssignmentService(db).update_assignment(assignment, update_schema)
 
-    await LmsSyncService(db).upsync_assignment(assignment)
```
ptlharit2, line 50: why are we deleting the upsync here? dont we use this endpoint to update the assignment dates back to canvas?

ptlharit2: ah yes I see its in the service now, ignore this

PR #77, opened 2024-08-27
add concurrent student sync, instructor sync, assignment sync. sync students and instructors simultaneously

frostyfan109:


app/services/lms_sync_service.py
```diff
@@ -86,6 +99,191 @@ async def sync_assignments(self):
...
+            pid, email, name = canvas_user.get("sis_user_id"), canvas_user.get("email"), canvas_user.get("name")
+            if pid is None or email is None or name is None:
+                print("Skipping over pending user", name or email or "<unknown>")
+                return None
+
+            try:
+                print("getting user info for ", pid)
+                user_info = self.ldap_service.get_user_info(pid)
+                print(pid, "->", user_info.onyen)
+            except:
```
Hoid: I think this should probably be an 
```
except UserNotFoundException:
    print("Skipping over student not in LDAP: ", pid or email or "<unknown>")
except:
    print("Some other error occurred")
```
kind of thing because currently you're swallowing all kinds of exceptions that the LDAP search could throw. If we run into a problem later where LDAP is failing consistently but we're seeing "Skipping over student not in LDAP:" in the logs and can't figure out why for a while that would suck.

frostyfan109: yes, that's true. I only copied the existing logic, which used a catch-all, but I can change it.

app/services/lms_sync_service.py
```diff
@@ -86,6 +99,191 @@ async def sync_assignments(self):
...
+            await self.canvas_service.associate_pid_to_user(resource_pair.user_info.onyen, resource_pair.user_info.pid)
+
+        if db_student is not None and canvas_student is not None:
+            # update
+            pass
+
+        if db_student is not None and canvas_student is None:
+            # delete
+            await self.student_service.delete_user(db_student.onyen)
+            try: await self.canvas_service.unassociate_pid_from_user(db_student.onyen)
```
ptlharit2: do we have something in the clean up service to handle this? 
If not, we can move this logic over into a single method to handle both student/instructor clean up - at a later date not important rn

ptlharit2 (approved):

app/services/lms_sync_service.py
frostyfan109: Handle what?

frostyfan109: Note: pid association is cascade deleted, the try/catch on the unassociate call can be deleted now

ptlharit2:
minor merge conflict merging develop into your branch with imports in `lms_sync_service` - just resolved it here while I was reviewing the PR

frostyfan109:
Locking issue occurs when trying to concurrently commit. Probably problematic in other areas as well...

```
PushRejected Error: exit status 1 - remote: * Checking 1 references        
remote: Checked 1 references in total        
remote: error: cannot lock ref 'refs/heads/main': is at 3aa5591dad0c73aac749626ed3c43afbca3b009e but expected 681fd19ebb8bb9537b422392da852aac4551a8b1        
To /data/git/gitea-repositories/eduhelx_spring_2024-dev-instructors/eduhelx_spring_2024-dev-class-master-repo.git
 ! [remote rejected] e87bd1bafa82b74df42f25b36e6aec35667fc598 -> main (failed to update ref)
error: failed to push some refs to '/data/git/gitea-repositories/eduhelx_spring_2024-dev-instructors/eduhelx_spring_2024-dev-class-master-repo.git'
 - remote: * Checking 1 references        
remote: Checked 1 references in total        
remote: error: cannot lock ref 'refs/heads/main': is at 3aa5591dad0c73aac749626ed3c43afbca3b009e but expected 681fd19ebb8bb9537b422392da852aac4551a8b1        
To /data/git/gitea-repositories/eduhelx_spring_2024-dev-instructors/eduhelx_spring_2024-dev-class-master-repo.git
 ! [remote rejected] e87bd1bafa82b74df42f25b36e6aec35667fc598 -> main (failed to update ref)
error: failed to push some refs to '/data/git/gitea-repositories/eduhelx_spring_2024-dev-instructors/eduhelx_spring_2024-dev-class-master-repo.git'
: remote: * Checking 1 references        
remote: Checked 1 references in total        
remote: error: cannot lock ref 'refs/heads/main': is at 3aa5591dad0c73aac749626ed3c43afbca3b009e but expected 681fd19ebb8bb9537b422392da852aac4551a8b1        
To /data/git/gitea-repositories/eduhelx_spring_2024-dev-instructors/eduhelx_spring_2024-dev-class-master-repo.git
 ! [remote rejected] e87bd1bafa82b74df42f25b36e6aec35667fc598 -> main (failed to update ref)
error: failed to push some refs to '/data/git/gitea-repositories/eduhelx_spring_2024-dev-instructors/eduhelx_spring_2024-dev-class-master-repo.git'
```

app/services/lms_sync_service.py
```diff
@@ -1,28 +1,41 @@
...
-from app.schemas.assignment import UpdateAssignmentSchema
+from app.services import CourseService, CanvasService, LDAPService, AssignmentService, StudentService, InstructorService, UserService, UpdateCanvasAssignmentBody, LDAPUserInfoSchema
+from app.models import AssignmentModel, StudentModel, SubmissionModel, InstructorModel, UserModel
+from app.schemas import UpdateCourseSchema, UpdateAssignmentSchema
 from app.core.exceptions import (
     AssignmentNotFoundException, NoCourseExistsException, 
     UserNotFoundException, LMSUserNotFoundException
 )
 
+class ResourcePair(BaseModel):
```
Hoid: I think this needs a more descriptive name. It's too generic rn, like if you want a different ResourcePair later on you'd have to rename this one.

Hoid: maybe DbLmsResourcePair?

app/services/lms_sync_service.py
```diff
@@ -91,6 +104,189 @@ async def sync_assignments(self):
...
         
         return canvas_assignments
+    
+    async def _pair_db_canvas_users(self, db_users: list[UserModel], canvas_users: list[dict], canvas_user_pids: list[str]):
+        # User exists in Canvas but not DB = create
+        # User exists in DB but not Canvas = delete
+        # User exists in both DB and Canvas = update
+        pairs = []
+        
+        async def process_canvas_user(canvas_user) -> ResourcePair | None:
```
Hoid: I think this should be called `create_resource_pair()` instead. "Processing" the canvas user doesn't mean anything, really.

app/services/lms_sync_service.py
```diff
@@ -91,6 +104,189 @@ async def sync_assignments(self):
...
+                user_info = self.ldap_service.get_user_info(pid)
+                print(pid, "->", user_info.onyen)
+            except UserNotFoundException:
+                print("Skipping over student not in LDAP: ", pid or email or "<unknown>")
+                return None
+
+            try:
+                user = await self.user_service.get_user_by_onyen(user_info.onyen)
+                return ResourcePair(db_resource=user, lms_resource=canvas_user, user_info=user_info)
+
```
Hoid, line 110: I would take out this whitespace line, but that's just a small thing

app/services/lms_sync_service.py
frostyfan109: It's not used anywhere besides this module, and in the context of the syncing module, I think it's named appropriately. If you feel strongly I'll rename

app/services/lms_sync_service.py
frostyfan109: It's named as such because it's an internal function used to process each canvas user (as opposed to the for block underneath which "processes" each database user. I can rename it but there was intent behind the name

frostyfan109: To be fair, I'm not sure why I wrote it in this way. It should be directly appending rather than returning to be honest, then the name would be more suitable.

app/services/lms_sync_service.py
Hoid: It's fine

app/services/lms_sync_service.py
Hoid: Right but the verb "process" doesn't really indicate what's happening here. Whereas "creating" a certain data type does. I just tend not to like vague verbs in function names ya know

app/services/lms_sync_service.py
```diff
@@ -91,6 +104,197 @@ async def sync_assignments(self):
...
+            pid = await self.canvas_service.get_pid_from_onyen(db_user.onyen)
+            if pid not in canvas_user_pids:
+                pairs.append(DatabaseLMSResourcePair(db_resource=db_user, lms_resource=None))
+        
+        # Then we process Canvas users
+        pairs += await asyncio.gather(*[process_canvas_user(u) for u in canvas_users])
+
+        return [p for p in pairs if p is not None]
+    
+    async def _pair_db_canvas_assignments(self, db_assignments: list[AssignmentModel], canvas_assignments: list[dict]):
```
Hoid: I feel like this `dict` should probably be a Model instead of just a dict

frostyfan109: You can type the canvas api interfaces if you'd like, but that would take a long time.

app/services/lms_sync_service.py
frostyfan109: The issue is the name you're suggesting is also misleading, because that function is not just creating resource pairs. It specifically is processing canvas users to derive update/create pairs.

app/services/lms_sync_service.py
Hoid: Why can't you just create a type that has the fields you want to pull in from the response, and drop the fields that you don't care about?

Hoid: I can do that, if you want

frostyfan109: Sure

app/api/api_v1/endpoints/student_router.py
```diff
@@ -33,17 +33,17 @@ async def list_students(
...
-    *,
-    db: Session = Depends(get_db),
-    perm: None = Depends(PermissionDependency(StudentCreatePermission)),
-    student_body: CreateStudentBody
-):
-    student = await StudentService(db).create_student(
-        **student_body.dict()
-    )
-    return student
+# @router.post("/students", response_model=StudentSchema)
```
Hoid, line 36: Is this temporary in the PR? If you're planning to commit this commented out, can you add another comment above saying why it's commented out?

Hoid: Is this temporary in the PR? If not can you add another comment above explaining why it's commented out?

frostyfan109: I'm not sure why it's commented I meant to delete it. This endpoint makes no sense given canvas integration, there is no "upsyncing" a user into the canvas course. The user will just get deleted upon the next down sync. Maybe I commented it in the case we add that? Idk

frostyfan109: Would you prefer a comment or just deleting the entire block ?

app/core/exceptions/user.py
```diff
@@ -1,7 +1,7 @@
 from .base import CustomException
 
 class UserAlreadyExistsException(CustomException):
-    code = 409
+    code = 400
```
Hoid: I prefer using 409 here, but if you want to change it to 400 I think that's fine we should just delineate between `UserAlreadyExistsException` and `DataConflictException` or something like that. The `DataConflictException` would be the one that returns 409 and would be returned if the client tries to _change_ a resource in a way that's incompatible, such as updating a submission to be open before the published date, or once we have a job queue, I think it'd be useful to return a 409 if the client tries to submit another request for a job where one is already in the queue (and only one is allowed at a time).

But my vote is just to do all of the above and also keep this to be a 409 because a `UserAlreadyExistsException` to me feels like a subclass of the `DataConflictException`.

Hoid: In general I think using specific error codes for (correctly applicable) specific scenarios is a good thing so that the client can respond differently to the different error codes. It helps the client respond differently if they know they got a 409 than if they got just a 400 for some other reason (which probably indicates a malformed request, not a data conflict).

frostyfan109:
I think I saw a lot of 400s elsewhere in user exceptiosns and wanted to make it consistent, but honestly 409 is probably more appropriate. I need to review our 400 vs 409 usage a bit more thoroughly, since although interchangeable don’t serve the same purpose semantically

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

app/core/exceptions/user.py
frostyfan109: I think I saw a lot of 400s elsewhere in user exceptions and wanted to make it consistent, but honestly 409 is probably more appropriate. I need to review our 400 vs 409 usage a bit more thoroughly, since although interchangeable don’t serve the same purpose semantically

frostyfan109: In reality, if I understand the nuance between 400 and 409 correctly, there really aren't any custom exceptions we should a 400 over 409 for. They should all change from 400 to 409 actually. That being said, I don't fully understand the nuance between them. I've always gotten conflicting information about their usage. Regardless, we should do a review to make sure our use is consistent across all the errors we have.

app/api/api_v1/endpoints/student_router.py
Hoid: I'd delete it if we don't have an explicit plan to use it later. The comment above a block of commented out code is just something we should do when we can't delete it because we plan to use it.

ptlharit2:
> Locking issue occurs when trying to concurrently commit. Probably problematic in other areas as well...
> 
> ```
> PushRejected Error: exit status 1 - remote: * Checking 1 references        
> remote: Checked 1 references in total        
> remote: error: cannot lock ref 'refs/heads/main': is at 3aa5591dad0c73aac749626ed3c43afbca3b009e but expected 681fd19ebb8bb9537b422392da852aac4551a8b1        
> To /data/git/gitea-repositories/eduhelx_spring_2024-dev-instructors/eduhelx_spring_2024-dev-class-master-repo.git
>  ! [remote rejected] e87bd1bafa82b74df42f25b36e6aec35667fc598 -> main (failed to update ref)
> error: failed to push some refs to '/data/git/gitea-repositories/eduhelx_spring_2024-dev-instructors/eduhelx_spring_2024-dev-class-master-repo.git'
>  - remote: * Checking 1 references        
> remote: Checked 1 references in total        
> remote: error: cannot lock ref 'refs/heads/main': is at 3aa5591dad0c73aac749626ed3c43afbca3b009e but expected 681fd19ebb8bb9537b422392da852aac4551a8b1        
> To /data/git/gitea-repositories/eduhelx_spring_2024-dev-instructors/eduhelx_spring_2024-dev-class-master-repo.git
>  ! [remote rejected] e87bd1bafa82b74df42f25b36e6aec35667fc598 -> main (failed to update ref)
> error: failed to push some refs to '/data/git/gitea-repositories/eduhelx_spring_2024-dev-instructors/eduhelx_spring_2024-dev-class-master-repo.git'
> : remote: * Checking 1 references        
> remote: Checked 1 references in total        
> remote: error: cannot lock ref 'refs/heads/main': is at 3aa5591dad0c73aac749626ed3c43afbca3b009e but expected 681fd19ebb8bb9537b422392da852aac4551a8b1        
> To /data/git/gitea-repositories/eduhelx_spring_2024-dev-instructors/eduhelx_spring_2024-dev-class-master-repo.git
>  ! [remote rejected] e87bd1bafa82b74df42f25b36e6aec35667fc598 -> main (failed to update ref)
> error: failed to push some refs to '/data/git/gitea-repositories/eduhelx_spring_2024-dev-instructors/eduhelx_spring_2024-dev-class-master-repo.git'
> ```

Is this locking issue still happening?

app/services/lms_sync_service.py
frostyfan109: Now that this method is a genuine "processor," i.e., it is processing canvas users into the pair list, I'm going to mark this resolved

frostyfan109:
> Is this locking issue still happening?

Since we're batching our repo modifications into a single commit there should be no locking anymore.

app/services/assignment_service.py
```diff
@@ -21,133 +25,150 @@
...
+                await self.get_assignment_by_id(assignment.id)
+                raise AssignmentAlreadyExistsException()
+            except AssignmentNotFoundException:
+                pass
+
+            try:
+                await self.get_assignment_by_name(assignment.name)
+                raise AssignmentAlreadyExistsException()
+            except AssignmentNotFoundException:
+                pass
```
Hoid: Why do we need both of these?

Hoid: Actually, why do we need either? I don't think the `createAssignments()` method should be doing implicit validation which requires doing database calls. This method should just be doing what it needs to do to create the actual assignments. There could be another method on the assignment_service that will check the db for if you can create the assignments, but I don't think the `create_assignments()` method should do that implicitly.

frostyfan109: 1. We need to validate every unique field, so both id and name. It needs to be two separate blocks; otherwise, consider if an assignment had a unique ID but non-unique name. It would throw before validating the name. I agree it could be a separate method.
2. I don't agree with separating creation from validation. It makes more sense in my opinion to raise an "assignment name taken" exception which integrates directly into our FastAPI error handling, than, say, a database integrity violation (which is what you would otherwise get from the method).

frostyfan109: At the end of the day though, assuming correct usage, there isn't any functional difference when decoupling validation. The only actual difference is assuming misuse (since you could now hit SQLAlchemy errors where not possible before). I'm not entirely opposed to decoupling; it does have its advantages like greater flexibility. On the other hand, it would require a refactor of every service that uses validation logic. Let me know what you wanna do

Hoid: > We need to validate every unique field, so both id and name. It needs to be two separate blocks; otherwise, consider if an assignment had a unique ID but non-unique name. It would throw before validating the name. I agree it could be a separate method.

That sounds like a data model problem to me. Just mark the column as UNIQUE then, and if you try and insert something that isn't unique and get an exception back, _then_ you deal with it that way.

Hoid: > I don't agree with separating creation from validation. It makes more sense in my opinion to raise an "assignment name taken" exception which integrates directly into our FastAPI error handling, than, say, a database integrity violation (which is what you would otherwise get from the method).

You don't have to raise a database integrity violation all the way up to FastAPI, you'd catch the database integrity violation and throw your own exception based on what you got. 

I reaaaally think we should separate creation from validation though. We can talk more about why offline.

frostyfan109: 1. The columns are unique... That's why you would get an integrity violation.
2. If you are catching the exception and reraising something more appropriate (NameAlreadyTaken) you are literally just reinventing the already present validation

frostyfan109: maybe I'm a bit confused why you feel separation of concerns here is so important?

Hoid: > If you are catching the exception and reraising something more appropriate (NameAlreadyTaken) you are literally just reinventing the already present validation

How are you "reinventing the already present validation"? All I'm saying is that you can catch a database exception and return a FastAPI exception so that the application fails in a graceful way.

> maybe I'm a bit confused why you feel separation of concerns here is so important?

I think in _this instance_ it's just important to separate creation from validation because the validation is requiring doing database calls for every assignment that you pass in to the method. If we can validate all of the assignments in one database call at the beginning that would be better.

Or better yet, I think we should be doing [upserts](https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#postgresql-insert-on-conflict) instead of using `add_all()`. Basically just attempt to insert the assignments into the table, but use the `on_conflict_do_nothing()` or `on_conflict_do_update()` method to define what should happen when the assignment already exists in the db.

Hoid (approved):
You'll need to fix the merge conflicts, but other than that I think this is good

PR #78, opened 2024-08-27
Manual grading

frostyfan109:


app/api/api_v1/endpoints/assignment_router.py
```diff
@@ -26,11 +29,21 @@ class UpdateAssignmentBody(BaseModel):
...
 
-class GradingBody(BaseModel):
+class OtterGradingBody(BaseModel):
     master_notebook_content: str
     otter_config_content: str
 
+class ManualGrade(BaseModel):
+    submission_id: int
+    # Between [0,1]
+    grade_proportion: float
```
Hoid, line 41: What is this proportion? Does it mean how much the grade is weighted in the class towards the final grade, or?

frostyfan109: It's the student's grade for the submission as a proportion

frostyfan109: It was renamed from grade percent after harit misinterpreted grade percent to be in [0,100].

Hoid: Oohh this is manual grading, that's right, my bad. Okay this makes sense.

app/api/api_v1/endpoints/assignment_router.py
```diff
@@ -89,13 +102,41 @@ async def grade_assignment(
...
         grading_body.otter_config_content
     )
-    return grade_report
\ No newline at end of file
+
+@router.post(
+    "/assignments/{assignment_name}/grade_manual",
+    response_model=GradeReportSchema
+)
+async def grade_assignment(
```
Hoid: This should be named `grade_assignment_manually()` not just `grade_assignment()` since the endpoint is `.../grade_manual`

app/services/submission_service.py
```diff
@@ -40,6 +47,14 @@ async def create_submission(
         self.session.add(submission)
         self.session.commit()
 
+        student_notebook_upload = await grading_service.get_student_notebook_upload(submission, student_notebook_content.encode())
```
Hoid: I'm worried that if there's an error thrown in the stuff you've added the stuff done previously needs to be rolled back. Maybe move this part to below the code you added?

```
self.session.add(submission)
self.session.commit()
```

frostyfan109: this change is done in submission downsync. I'm not particularly sure how the syncing change above got included in manual-grading and submission-downsync but not in submission-upload-cleanup which is the parent of both...

app/core/exceptions/grading.py
```diff
@@ -3,4 +3,19 @@
...
+
+class AutogradingDisabledException(CustomException):
+    code = 400
+    error_code = "GRADING__AUTOGRADING_DISABLED"
+    message = "autograding has been disabled for the assignment"
+
+class StudentGradedMultipleTimesException(CustomException):
+    code = 400
+    error_code = "GRADING__STUDENT_GRADED_MULTIPLE_TIMES"
+    message = "only 1 submission per student is permitted for batch grading"
```
Hoid: I would use a term here other than "batch grading" if you expect that this error might ever be shown to users. It's kind of an esoteric term for the lay person, and the message was kind of confusing for me too at first. What about "only one submission per student is permitted during grading"? We don't have to stipulate "batch grading" at all since the scenario where we grade only one submission total precludes the possibility of attempting to grade multiple submissions from one student.

app/services/canvas_service.py
```diff
@@ -339,61 +335,70 @@ async def create_course_folder(
...
+    Note that Canvas does not support associating grades with submissions.
+    You can only set the overall assignment grade for the student.
+    Comments will be attached to the active submission.
+    """
+    async def upload_assignment_grade(
+        self,
+        assignment_id: int,
+        user_id: int,
+        # between [0,1]
+        grade_proportion: float,
```
Hoid, line 385: The comment above this says it can be any value between 0 and 1. Why not enforce that with a type? Maybe like this?
```python
class GradeProportion(float):
    def __new__(cls, input = 0, *args, **kwargs):
        instance = int.__new__(cls, input, *args, **kwargs)
        if not 0 <= instance <= 1:
            raise ValueError('Value outside GradeProportion range')
        return instance
```

Hoid: Then you can only accept a `GradeProportion` here instead of any float, and you won't have to add validation within this method.

frostyfan109: Oh yes there's a way to do that with pydantic I think

frostyfan109: It's confloat with ge=0, le=1

Hoid: Looks like its use is discouraged in the docs, since it doesn't play well with static analysis tools: https://docs.pydantic.dev/latest/api/types/#pydantic.types.confloat

They give another solution, to use this
```python
class Foo(BaseModel):
    bar: Annotated[float, Field(strict=True, gt=0)]
```
which I thiiink works just as well as the class I gave above. I do like that the class implementation is named, though, rather than the field just having properties that you don't know of in the code unless you look on the base model.

frostyfan109: An annotated float works just as well. I would keep in mind we are using I believe pydantic v1.

app/services/grading_service.py
```diff
@@ -225,4 +241,78 @@ async def grade_assignment(
             # All we've done is change `graded` on submissions, which can't cause any violations here.
             self.session.commit()
             
-            return grade_report
\ No newline at end of file
+            return grade_report
+        
+    async def grade_assignment_manual(
```
Hoid: Can this be named `grade_assignment_manually()`? lol If you care to keep it the same it's fine it's just a preference I have that you might agree with 🤷‍♂️

frostyfan109: Yeah I don't particularly care

app/services/canvas_service.py
Hoid: Sounds good. So are you going to change this float to an annotated float with constraints, or leave it as is?

app/services/lms_sync_service.py
```diff
@@ -191,12 +194,32 @@ async def sync_instructors(self):
...
+        submission: SubmissionModel,
+        student_notebook_content: bytes
+    ):
+        user_pid = await self.canvas_service.get_pid_from_onyen(submission.student.onyen)
+        
+        # If this course runs on a 2U Digital Campus instance, append ":UNC" to the PID
+        if "digitalcampus" in settings.CANVAS_API_URL:
+            user_pid += ":UNC"
+
+        student = await self.canvas_service.get_student_by_pid(user_pid)
```
Hoid, line 203: Is this just done as validation? Don't we already have the student's information under `submission.student`?

frostyfan109: that's the canvas student, not the database student. i named it as such to keep in line with elsewhere in the service.

app/services/canvas_service.py
frostyfan109: yes i will test it

app/services/submission_service.py
Hoid: So...what needs to happen here? Can we make this code exactly the way it should be for now?

app/services/lms_sync_service.py
Hoid: I think it'd be better to just change them all to be named `canvas_student`, which is more clear considering there's a database student being used up above in a lot of these methods, instead of just acquiescing to the norm.

Hoid:
Okay I've reviewed all the code and other than the outstanding convos in this PR I think it's all good. Just need to resolve those convos and I think it's good to merge.

app/services/submission_service.py
frostyfan109: This code is outdated, refer to latest version of it.

frostyfan109: In fairness, even after merging in the initial submission sync, the actual cleanup code seems to live on the second "submission downsync" PR, so the issue still persists it's just been relocated.

app/services/lms_sync_service.py
Hoid: Actually nevermind I think this one is fine, don't want to hold up the PR for it

Hoid (approved):

PR #83, opened 2024-10-29
Job queue

frostyfan109:


app/core/config.py
```diff
@@ -91,6 +100,24 @@ def compute_instructor_appstore_api_url(cls, v: Optional[str], values: Dict[str,
         if isinstance(v, str): return v
         return values.get("INSTRUCTOR_APPSTORE_HOST") + "/api/v1"
 
+    @validator("CELERY_BROKER_URI", pre=True)
+    def assemble_broker_uri(cls, v: Optional[str], values: Dict[str, Any]) -> Any:
```
Hoid: Shouldn't this one and the function below it return a `str`?

app/celery/worker.py
```diff
@@ -0,0 +1,60 @@
...
+    worker_enable_remote_control=True,
+    beat_schedule={
+        "periodic-downsync": {
+            "task": "downsync",
+            "schedule": 60
+        }
+    }
+)
+
+def get_tasks_by_name(
```
Hoid: This should be called `get_task_ids_by_name()`

app/core/config.py
frostyfan109: I guess; it's inconsequential but will change it.

app/celery/worker.py
```diff
@@ -0,0 +1,60 @@
...
+    if include_pending:
+        for item in redis_client.lrange("celery", 0, -1):
+            task_data = json.loads(item)
+            if task_data.get("headers", {}).get("task") == task_name:
+                task_id = task_data["headers"].get("id")
+                if task_id is not None: task_ids.append(task_id)
+
+    return task_ids
+
+def get_task_status_by_id(task_id: str) -> AsyncResult:
```
Hoid: This isn't really getting the task status, it's getting the whole returned value of `AsyncResult`. You'd have to call `ready()` on it to get the status. I think this should either be renamed to `get_task_result_by_id()` or changed to call `ready()` on the AsyncResult before returning.

app/api/api_v1/endpoints/lms_router.py
```diff
@@ -17,13 +20,14 @@ class GradeUpload(BaseModel):
...
-@router.post("/lms/downsync")
+@router.post("/lms/downsync", response_model=UUID)
 async def downsync(
     *,
     db: Session = Depends(get_db),
     perm: None = Depends(PermissionDependency(UserIsInstructorPermission))
 ):
-    await LmsSyncService(db).downsync()
+    task = downsync_task.delay()
+    return task.id
```
Hoid: Will this return object have a key at all, or will it just be the UUID itself? If the latter, I feel like it should return with a key called "task_id" or something and the value is the UUID.

frostyfan109: what?

app/api/api_v1/endpoints/job_router.py
```diff
@@ -0,0 +1,23 @@
...
+
+router = APIRouter()
+
+@router.get("/jobs/{id}", response_model=JobSchema)
+async def get_job(
+    *,
+    id: UUID,
+    perm: None = Depends(PermissionDependency(UserIsInstructorPermission))
+):
+    return JobSchema.from_async_result(get_task_status_by_id(str(id)))
```
Hoid: I don't think we need to expose the whole job return value of `AsyncResult` to an endpoint, personally. That's a lot of info that's kind of superfluous. If you end up changing `get_task_status_by_id()` to call `ready()` on the AsyncResult then that's fine, and I would just change this method to be called `get_job_status()` and change the endpoint to be `/jobs/{id}/status`.

.env.sample
```diff
@@ -27,6 +35,8 @@ GITEA_ASSIST_AUTH_TOKEN="YOUR_BEARER_TOKEN"
 CANVAS_API_KEY="YOUR_API_KEY"
 CANVAS_API_URL="https://uncch.instructure.com/api/v1"
 CANVAS_COURSE_ID="12345"
+CANVAS_COURSE_START_DATE="1970-01-01T00:00:00Z"
```
Hoid, line 38: Currently in my .env.sample file, I have this variable set to this format for a date: `CANVAS_COURSE_START_DATE="06/26/2024 00:00:00"`. Are you sure the format you're using works? If so that's cool, just wanna make sure.

app/api/api_v1/endpoints/lms_router.py
Hoid: Like will the response body be 
```
{ 92F8f118ef929jf-f19j1f-f19mf2-1f91 }
```
or 
```
{ 
  "task_id": "92F8f118ef929jf-f19j1f-f19mf2-1f91"
}
```
I think it should be the latter.

start.py
```diff
@@ -48,15 +48,6 @@ def main(host: str, port: int, reload: bool, workers: int | None=None):
     alembic_cfg = Config("alembic.ini")
     command.upgrade(alembic_cfg, "head")
 
-
-    # Run setup wizard, if required.
-    try:
```
Hoid, line 54: You've removed the `downsync` run from `start.py`, but haven't added it back somewhere using the workers. Is that intentional? I would think we would want to run the downsync before starting the app like we usually do.

app/api/api_v1/endpoints/lms_router.py
frostyfan109: no the response body is a string?

.env.sample
frostyfan109: yes it's an iso formatted datetime. it's possible that other formats are supported by pydantic deserialization.

start.py
frostyfan109: it has been added back into worker.py.

app/api/api_v1/endpoints/job_router.py
frostyfan109: added /jobs/{id} for full result status, /jobs/{id}/status indicating only {id, status}.

app/api/api_v1/endpoints/lms_router.py
frostyfan109: changed to return a JobSchema of id and status.

README.md
```diff
@@ -23,11 +24,16 @@ vim .env
...
 set -a && source .env
+python start.py -r
+# or
 python -m app.main
 # or
 uvicorn --reload app.main:app --log-level=info
 ```
 
+It is assumed that you've configured your .env properly. This includes the expectation that a Postgres
+instance, Gitea Assist, message broker, and celery worker(s) are all configured and running.
```
ptlharit2: Any message broker works or specifically redis? `assemble_broker_uri` in config.py are creating redis specific URIs so not sure

app/api/api_v1/endpoints/lms_router.py
```diff
@@ -17,13 +20,14 @@ class GradeUpload(BaseModel):
...
 
-@router.post("/lms/downsync")
+@router.post("/lms/downsync", response_model=JobSchema)
 async def downsync(
     *,
     db: Session = Depends(get_db),
     perm: None = Depends(PermissionDependency(UserIsInstructorPermission))
 ):
-    await LmsSyncService(db).downsync()
+    task = downsync_task.delay()
```
ptlharit2, line 29: i'm guessing `delay()` is a celery specific method, for my knowledge, why use it here?

app/celery/worker.py
```diff
@@ -0,0 +1,60 @@
...
+                task_id = task_data["headers"].get("id")
+                if task_id is not None: task_ids.append(task_id)
+
+    return task_ids
+
+def get_task_result_by_id(task_id: str) -> AsyncResult:
+    return celery_app.AsyncResult(task_id)
+
+def cancel_task_by_id(task_id: str):
+    celery_app.control.revoke(task_id, terminate=True, signal="SIGKILL")
```
ptlharit2, line 69: is it possible to return a confirmation status of the task being cancel - though I am not sure if it's entirely needed

README.md
frostyfan109: Yeah, I'll update the text. We're going to go with strictly Redis.

app/api/api_v1/endpoints/lms_router.py
frostyfan109: delay and apply_async are methods by which you can dispatch a task. delay is a shorthand.

app/celery/worker.py
frostyfan109: what? if you mean adding extra logic to the function to confirm the task is cancelled, no, it is not practical.

app/celery/signals.py
```diff
@@ -0,0 +1,72 @@
...
+                              task_kwargs=unique_kwargs,
+                              key_prefix=app_config.key_prefix)
+
+    # clean lock
+    cache_task_id = backend.get(redis_key)
+    if cache_task_id and cache_task_id.startswith(task_id):
+        print(f"Cleaning singletion lock: {redis_key}, task_id: {task_id}")
+        backend.clear(redis_key)
+
+# @task_received.connect
```
Hoid, line 54: Can we either delete all this commented out code or add a comment above it describing why it's commented out?

frostyfan109: it is to be integrated with websockets which will build off the PR. I don't think it's worth a commit to remove/add comment when it's about to be uncommented

Hoid (approved):

ptlharit2 (approved):

frostyfan109:
Websocket-pubsub supercedes

PR #84, opened 2024-11-01
Fix logging middleware and add file logger

Hoid:
* Disable other loggers like the root logger and `uvicorn.access` so that it doesn't log every API request twice
* Add request ID to every request; this could eventually be propagated through Celery so that each job has that request ID associated with it and we can trace which request created the job and when looking through logs after the fact can correlate logs in workers to specific requests
* Remove the response body from the stdout logs, but keep the request body
* Add a file logger so that the full request with the response body is logged in a rotating file on disk (in /app/logs/debug.log)
* Add health endpoints so that we can stop calling /docs for the liveness and readiness probes

### DIFF
--- a/app/api/api_v1/endpoints/health_router.py
+++ b/app/api/api_v1/endpoints/health_router.py
@@ -1,0 +1,13 @@
+from fastapi import APIRouter
+
+
+router = APIRouter()
+
+@router.get("/health/live", response_model=None, status_code=200)
+async def heathcheckLive():
+  pass
+
+@router.get("/health/ready", response_model=None, status_code=200)
+async def heathcheckReady():
+  # In the future, check the health of redis and other internal systems here
+  pass
--- a/app/api/api_v1/endpoints/health_router.py
+++ b/app/api/api_v1/endpoints/health_router.py
@@ -4,10 +4,10 @@ from fastapi import APIRouter
 router = APIRouter()
 
 @router.get("/health/live", response_model=None, status_code=200)
-async def heathcheckLive():
+async def healthcheck_live():
   pass
 
 @router.get("/health/ready", response_model=None, status_code=200)
-async def heathcheckReady():
+async def healthcheck_ready():
   # In the future, check the health of redis and other internal systems here
   pass
--- a/app/core/middleware/logger.py
+++ b/app/core/middleware/logger.py
@@ -1,59 +1,130 @@
 import json
+import logging
 import time
+from typing import Callable
+from uuid import uuid4
+from fastapi import FastAPI, Response
 from starlette.types import Message
 from starlette.requests import Request
 from starlette.middleware.base import BaseHTTPMiddleware
-from app.core.middleware.iterator_wrapper import iterator_wrapper as aiwrap
+from app.core.utils.async_iterator_wrapper import AsyncIteratorWrapper
 
 class LogMiddleware(BaseHTTPMiddleware):
 
-    # Leaving the code in here that sets up the request and response body just in
-    # case we want to use it in the future. But for now it's taking up way too much
-    # log space for no benefit.
+    def __init__(self, app: FastAPI, *, logger) -> None:
+        self._logger = logger
+        super().__init__(app)
 
     # https://github.com/tiangolo/fastapi/issues/394#issuecomment-927272627
-    # async def set_body(self, request: Request):
-    #     receive_ = await request._receive()
+    async def set_body(self, request: Request):
+        receive_ = await request._receive()
 
-    #     async def receive() -> Message:
-    #         return receive_
+        async def receive() -> Message:
+            return receive_
 
-    #     request._receive = receive
+        request._receive = receive
 
     async def dispatch(self, request, call_next):
-            start_time = time.time()
+        request_id: str = str(uuid4())
+        logging_dict = {
+            "X-API-REQUEST-ID": request_id  # X-API-REQUEST-ID maps each request-response to a unique ID
+        }
 
-            # await self.set_body(request)
-            # req_body = await request.body()
-            response = await call_next(request)
+        await self.set_body(request)
+        response, response_dict = await self._create_response_log(
+            call_next,
+            request,
+            request_id
+        )
+        request_dict = await self._create_request_log(request)
+        logging_dict["req"] = request_dict
+        logging_dict["res"] = response_dict
 
-            # Consuming FastAPI response and grabbing body here
-            # resp_body = [section async for section in response.__dict__['body_iterator']]
-            # Repairing FastAPI response
-            # response.__setattr__('body_iterator', aiwrap(resp_body))
+        # urlStr = str(request.url)
+        # if urlStr[-5:] != "/docs":
+            # self._logger.info(
+            #     {
+            #         "req": {
+            #             "method": request.method, 
+            #             "url": str(request.url),
+            #             # "body": req_body,
+            #             "user": request.user.onyen if hasattr(request.user, "onyen") else None,
+            #         },
+            #         "res": {
+            #             "status_code": response.status_code
+            #             # "response_time": f"{round((time.time() - start_time) * 1000)} ms",
+            #             # "body": resp_body
+            #         }
+            #     }
+            # )
+        self._logger.info(logging_dict)
 
-            # Formatting response body for logging
-            # try:
-            #     resp_body = json.loads(resp_body[0].decode())
-            # except:
-            #     resp_body = str(resp_body)
+        return response
+    
+    async def _create_request_log(self, request: Request) -> dict:
 
-            request.app.logger.info(
+        path = request.url.path
+        if request.query_params:
+            path += f"?{request.query_params}"
+
+        request_log_dict = {
+            "method": request.method
+        }
+
+        try:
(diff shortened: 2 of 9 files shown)


app/api/api_v1/endpoints/health_router.py
```diff
@@ -0,0 +1,13 @@
+from fastapi import APIRouter
+
+
+router = APIRouter()
+
+@router.get("/health/live", response_model=None, status_code=200)
+async def heathcheckLive():
```
frostyfan109: should be snake case

app/api/api_v1/endpoints/health_router.py
```diff
@@ -0,0 +1,13 @@
...
+  pass
+
+@router.get("/health/ready", response_model=None, status_code=200)
+async def heathcheckReady():
```
frostyfan109: camelCase?

app/core/middleware/logger.py
```diff
@@ -1,59 +1,125 @@
...
-            response = await call_next(request)
+        await self.set_body(request)
+        response, response_dict = await self._create_response_log(
+            call_next,
+            request,
+            request_id
+        )
+        request_dict = await self._create_request_log(request)
+        logging_dict["req"] = request_dict
+        logging_dict["res"] = response_dict
```
frostyfan109: you could just be setting this in the literal construct it seems like?

app/core/middleware/logger.py
```diff
@@ -1,59 +1,125 @@
...
+
+            self._file_logger.debug(logging_dict)
+
+        return response
+    
+    async def _create_request_log(self, request: Request) -> str:
+
+        path = request.url.path
+        if request.query_params:
+            path += f"?{request.query_params}"
```
frostyfan109: why does this exist?

app/core/middleware/logger.py
```diff
@@ -1,59 +1,125 @@
...
+
+        request_log_dict = {
+            "method": request.method,
+            "endpoint": request.url.path,
+            "user": request.user.onyen if hasattr(request.user, "onyen") else None,
+        }
+
+        try:
+            body = await request.json()
```
frostyfan109: unless you've cloned something here I'm pretty sure you can't consume the request body inside the logger or it's not going to be readable in the actual handler? have you tested?

app/main.py
```diff
@@ -64,27 +88,29 @@ def make_middleware() -> List[Middleware]:
...
-        openapi_url=f"{ settings.API_V1_STR }/openapi.json",
-        middleware=make_middleware()
+@app.exception_handler(CustomException)
+async def custom_exception_handler(request: Request, exc: CustomException):
+    content = { "error_code": exc.error_code, "message": exc.message }
+    if settings.DEV_PHASE == DevPhase.DEV:
+        content["stack"] = exc.stack
+    return JSONResponse(
+        status_code=exc.code,
+        content=content,
```
frostyfan109: why move this?

app/main.py
```diff
@@ -64,27 +88,29 @@ def make_middleware() -> List[Middleware]:
...
-    init_routers(app)
-    init_listeners(app)
-    add_pagination(app)
-    
-    return app
 
-app = create_app()
+init_monkeypatch()
+init_routers(app)
+add_pagination(app)
```
frostyfan109: why ?

app/core/middleware/logger.py
Hoid: not sure what you mean by this

app/core/middleware/logger.py
Hoid: So that query params are included in the log. I forgot to change line 73 to use `path` instead of `request.url.path`, so I can change that. Do you think it's not necessary?

app/core/middleware/logger.py
Hoid: There's a copy of the request body in the `_json` field on the request. `request.json()` just pulls that copy. I've tested it, the actual endpoint in the router has the full request body and can use it.

app/main.py
Hoid: Because the way it was done before meant the `custom_exception_handler`  function showed up as unused by the linter. Plus, it was unnecessary to have the method inside a wrapper method when we can just do it this way.

app/main.py
Hoid: Why put all this in a method and call that method right afterward instead of just doing it at the root level? I just thought this made more sense, wasn't sure why it was done the other way.

app/core/middleware/logger.py
Hoid: You can check out the logs in the grader-api in the `eduhelx-test` namespace to see how the logs look

app/core/middleware/logger.py
frostyfan109: You could set these fields where you declare the dictionary.

app/core/middleware/logger.py
frostyfan109: Path isn't used anywhere or am I missing something?

frostyfan109: No it's just not being used, yes.

Hoid: I'm not sure why it's not showing up in GH but it's used on line 73 instead of `request.url.path`

app/main.py
frostyfan109: The purpose really just boils down to organization. Particularly in the above where init_listeners was removed, which in the job queue PR adds multiple new listener signals to the app lifecycle.

frostyfan109: Either way, it's really inconsequential for the create app method. But it is specifically organized as such to be consistent with the organization/structuring of the module.

Hoid: I would think we would just put each annotated listener methods at the root level after we create the `app` variable. I can change it back if you prefer, just let me know if you think it's important or if this other way might be better.

app/main.py
```diff
@@ -10,25 +12,47 @@
...
+    logger.addHandler(stream_handler)
+    return logger
+
+def init_file_logger(formatter) -> logging.Logger | None:
+    file_logger = None
+    if settings.DEV_PHASE == DevPhase.PROD:
+        file_logger = logging.getLogger("file-logger")
+        file_logger.handlers.clear()
+        file_logger.setLevel(logging.DEBUG)
+        file_handler = RotatingFileHandler("logs/debug.log", maxBytes=1024*1024*100, backupCount=4)
```
ptlharit2, line 42: do we lose these log files if we kill the pod?

app/main.py
```diff
@@ -10,25 +12,47 @@
...
+formatter = logging.Formatter("%(asctime)s %(levelname)s %(message)s")
+
+def init_logging():
+    logging.getLogger("uvicorn.access").disabled = True
+    root_logger = logging.getLogger()
+    root_logger.disabled = True
+    for handler in root_logger.handlers[:]:
+        root_logger.removeHandler(handler)
+
+def init_uvicorn_logger(formatter) -> logging.Logger:
```
ptlharit2, line 28: I thought we were setting the log format through the json config file? that approach would make it more customizable and easier to swap between log styles. This is fine too but I thought we were already using the file to format

start.py
```diff
@@ -58,7 +59,7 @@ def main(host: str, port: int, reload: bool, workers: int | None=None):
         print(str(e))
 
     # Start the application
-    uvicorn.run("app.main:app", host=host, port=port, reload=reload, workers=workers)
+    uvicorn.run("app.main:app", host=host, port=port, reload=reload, workers=workers, log_config=None)
```
ptlharit2, line 62: this is where I thought we can pass the config, but again its just a bonus

app/main.py
frostyfan109: It's fine. I don't see any particular reason one way or another at the moment. If we find good reason to change it back, it's simple enough to do so at that point.

app/main.py
frostyfan109: I do think we need associated changes on the chart to store logs into a PVC. This file path should probably be controlled by some environment variable rather than fixed, I'm thinking.

ptlharit2: yeah that was my thought process, whether we'd lose the logs if the pod dies. It should be stored in a PVC but I wouldn't hold back this PR @Hoid - maybe just make a new jira ticket to track the work of persistent logs storage. We can expand on whether that PVC should hold other logs too, something to tackle as part of that ticket IMO

start.py
Hoid: If we pass a config in through here it messes things up, because it sets up the default logger. We want to disable the default logger (uvicorn, uvicorn.access, etc.) because those would double log everything but not in the format we want.

app/main.py
Hoid: We weren't setting the formatter in the logging_config.json, we were only setting the log level and the log retention. Now we're just using a rotating file handler so it once a log file gets to be large enough (~1MB) it will save as an archive backup and start logging to a new one, for up to 4 backups. And the log level could probably be changed based on `DEV_PHASE`, so I could do that if you want, but I don't think we need a logging_config.json anymore personally.

app/main.py
Hoid: I moved it back to the way it was

frostyfan109 (approved):

PR #85, opened 2024-11-12
Websocket pubsub

frostyfan109:
- Implements coordinated websocket messaging via pubsub channel
- Adapt existing auth middleware to support query param-based auth only for websocket handshakes
- Track job status updates in database table and emit as websocket messages
- Track internal crud events and emit as websocket messages
- Adds endpoint for fetching the current job id of the downsync job, if it is active/queued


Hoid:
it says I changed the base branch to `job-queue` instead of develop like 3 months ago, but I don't remember doing that nor why I would. is it correct the way it is or should we change it back to develop? @frostyfan109

frostyfan109:
> it says I changed the base branch to `job-queue` instead of develop like 3 months ago, but I don't remember doing that nor why I would. is it correct the way it is or should we change it back to develop? @frostyfan109

That PR was closed so this needs to merge into develop

README.md
```diff
@@ -23,11 +24,36 @@ vim .env
...
 ```bash
 set -a && source .env
+python start.py -r
+# or
 python -m app.main
 # or
 uvicorn --reload app.main:app --log-level=info
 ```
 
+It is assumed that you've configured your .env properly. This includes the expectation that a Postgres
```
Hoid, line 34: I think we should develop a docker-compose for local development soon that stands all these up as containers at once. What do you think @frostyfan109?

app/api/api_v1/endpoints/job_router.py
```diff
@@ -0,0 +1,39 @@
...
+
+@router.get("/jobs/{id}", response_model=JobResultSchema)
+async def get_job(
+    *,
+    id: UUID,
+    perm: None = Depends(PermissionDependency(UserIsInstructorPermission))
+):
+    """
+    NOTE: Task ID is not validated. A pending task does not necessarily exist;
+    this state is returned by default for unknown task ids.
```
Hoid, line 25: What does "this state" mean exactly?

README.md
frostyfan109: Yes, that's a good idea, I agree. I'll capture it in a ticket.

app/api/api_v1/endpoints/job_router.py
frostyfan109: Pending state

Hoid: I'm assuming that there's some underlying reason why this is the way it is, like Celery itself returns `Pending` for jobs that don't exist? If so, is it important to know when a task doesn't exist, or do we not really care if it exists or not if we query a job id and get `Pending` in most cases? I feel like this method should either check if the job exists before everything and return a different state if not, or if it gets a `Pending` state back from `JobResultSchema.from_async_result(get_task_result_by_id(str(id)))` it should then check to make sure the job exists.

app/celery/signals.py
```diff
@@ -0,0 +1,103 @@
...
+from app.database import SessionLocal
+from app.services import WebsocketManagerService, JobStatusService, InstructorService
+from app.models.job_status import JobStatusModel, JobStatus
+from app.schemas import JobSchema, WebsocketJobStatusMessage
+
+def handle_job_status_update(job_id: str, job_type: str | None, job_status: JobStatus) -> None:
+    """ Store a job status update in the database and dispatch a websocket event. """
+    async def _handle_job_status_update():
+        with SessionLocal() as session:
+            job_status_model = JobStatusService(session).create_job_status_update(job_id, job_type, job_status)
```
Hoid, line 20: Why not have the `JobStatusService` create the db session itself if the lifecycle of the service is within a specific method? I think the constructor of JobStatusService should accept a session as a parameter but default the value to `None`, and if it's `None` then it creates the session for itself. I think that's a common enough occurrence that it makes sense and both simplifies the process of, and makes clear that we are, creating the service in a way where the session should only live for the lifetime of the service/method. I think all services we have should use that pattern.

app/api/api_v1/endpoints/job_router.py
frostyfan109: Yes, that's what the comment is trying to convey. It's not important to know (and not particularly easy to do so using Celery directly without inspecting the broker store which I'm not convinced is even a consistent solution and is likely still somewhat ambiguous), nor should the method be doing so due to the complexity it introduces, without marked benefit. A non-existent task is implied to be pending, it's just the default state. If you're using this endpoint, then you should have a task id provided to you by the API. That means you can distinguish this ambiguity for yourself. I don't see any other realistic use case.

app/celery/signals.py
```diff
@@ -0,0 +1,103 @@
...
+            )
+            
+            if job_type == "downsync":
+                user_list = await InstructorService(session).list_instructors()
+            else:
+                user_list = []
+
+            WebsocketManagerService.publish_sync_pubsub_ws_message(ws_message, user_list)
+    
+    asyncio.run(_handle_job_status_update())
```
Hoid, line 32: What's the difference between this and just making the `handle_job_status_update()` method async? In the way you have it you're kind of disguising the fact that the internals of this function are async, which I don't love.

app/celery/signals.py
frostyfan109: Services should not be creating database sessions ever. They are intended to be stateless. If the service is instantiating a database session in its constructor, then there is no way to distinguish when to teardown such a session, resulting in pool overflow and a variety of other problems.

app/celery/signals.py
frostyfan109: It has been so long since I've done these changes that I can't say this will be accurate with 100% confidence, but I'm fairly certain I had to write it in such a way due to the way asynchronous code and the event loop is handled on Celery workers, where signals are triggered. If you did it the typical way (that you would expect to see here) it would cause errors related to asyncio.

Also, the function is not async, it is synchronous. That is what asyncio.run is doing, making it synchronous. Yes, it is a little unfortunate that the "internals" are executing in an asynchronous context which is not demonstrated by the signature, but that's not important to the function as a whole I don't think. If you think it's important, we can separate the functions out and rename them appropriately to indicate this better. Or, I could just rename the internally nested appropriately instead of simply underscoring it.

I'll add a comment explaining this, that was an oversight on my part.

app/celery/signals.py
Hoid: That's fair, but then shouldn't this function be stateless as well? We should be passing in the session from the caller instead of creating it here.

frostyfan109: I'm not sure what you mean. Why would that imply this function should also be stateless? I don't think there's any practical difference between instantiating a session context here vs in each individual signal handler.

Hoid: Actually better yet, I don't think we should be storing the status of a job in the db at all. I think we should be storing some sort of state of jobs in an in-memory store like redis.

frostyfan109: I guess we could, like with an event sourcing stream, but I did think about it a little when implementing this. I don't really see any decisive benefits to it. Sure, we don't necessarily need perpetual persistence of these statuses, but it's not as if it's particularly harmful, so not much of a benefit there. It would be theoretically offer faster querying, but that really doesn't matter at all in this context. We're also not storing simple key-values here, so I'm not even sure using redis streams to implement event sourcing would be viable (which would be the only realistic reason to be using redis for this) since I don't believe they support hashsets.

Is there a reason in particular you think it should be done this way?

frostyfan109: Just to clarify my position here, I don't think it's a bad idea. I think there could be a lot of utility in using Redis throughout our infrastructure to a greater extent than solely to support the job queue and pubsub functionality, so I'm not really opposed to building out the functionality to use Redis differently in areas like this.

But, I do think when we deviate from the primary relational database, we should have good reason to do so and be able to justify that design decision, since things like this introduce more complexity within the project and that's something I think we should seek to avoid.

app/celery/signals.py
```diff
@@ -0,0 +1,103 @@
...
+def task_received_handler(request, **kw):
+    handle_job_status_update(request.task_id, request.task_name, JobStatus.RECEIVED)
+
+@task_prerun.connect
+def task_prerun_handler(sender=None, task_id=None, **kw):
+    handle_job_status_update(task_id, sender.name, JobStatus.STARTED)
+
+@task_success.connect
+def task_success_handler(sender=None, result=None, **kw):
+    handle_job_status_update(sender.request.id, sender.name, JobStatus.SUCCESS)
```
Hoid, line 44: Is there a way we can use the `result` here, if it exists for a certain task? We might not be using this now for any certain tasks, but if we add a task eventually that outputs a result we'd want to be able to use it in some way right?

app/celery/signals.py
```diff
@@ -0,0 +1,103 @@
...
+
+@task_failure.connect
+def task_failure_handler(sender=None, task_id=None, **kw):
```
Hoid, line 47: There's an argument that can be passed into this function called `exception` that contains the exception that's raised. Will that exception be bubbled up where it will be properly logged, or do we need to log it here?

app/celery/signals.py
```diff
@@ -0,0 +1,103 @@
...
+    handle_job_status_update(task_id, sender.name, JobStatus.FAILURE)
+
+@task_internal_error.connect
+def task_internal_error_handler(sender=None, task_id=None, **kw):
```
Hoid, line 51: Same thing here for an exception that's raised. Do we need to log it here or do anything with it?

app/celery/signals.py
```diff
@@ -0,0 +1,103 @@
...
+
+""" Source: https://github.com/steinitzu/celery-singleton/issues/29 """
+@task_revoked.connect
+def clean_singleton_lock_on_revoke(sender=None, request=None, reason=None, **kwargs):
+    """ For some reason, Celery devs decided to remove basically every task workflow callback,
+    including on_revoked, which means that celery-singleton has no idea if a task is revoked.
+    We need to manally fix this here...
+    """
+    task_id = request.id
+    task_name = request.task
```
Hoid, line 70: the `request` parameter is a `celery.app.task.Context` object, not a `task.request`, so I don't think it has a `task` field. here's the class definition for `Context`: https://docs.celeryq.dev/en/latest/_modules/celery/app/task.html#Context

app/celery/signals.py
frostyfan109: I don't think the result is relevant here specifically, since the purpose of the signals as they are right now is for job _status_ tracking. There is an endpoint in the job router for retrieving the result of a task (`JobResultSchema`). There's isn't much merit in persisting the result of the task to this table, since it's pretty trivial to retrieve its result using Celery itself.

app/celery/signals.py
frostyfan109: It is bubbled up and logged automatically, so no need to do so here.

app/celery/signals.py
frostyfan109: Same thing here, it will be logged.

app/celery/signals.py
frostyfan109: I recall looking into this some when originally implementing it, because this note in the documentation also caught my attention. Regardless, it looks like I made a mistake anyways because as the documentation states, when using prefork pooling, the `task` field will not be defined. Perhaps this is a restriction we have to live with (no preforking), since we certainly need access to the underlying task function and I'm not sure how we would otherwise access it.

At the same time, the documentation implies that it both should be defined anyways and isn't:
image
I.e., it seems to state that a Context _is_ a `Task.request`; however, referencing the source code of Context, there is no `task` field. I am leaning towards it being defined--if you look into the source code, its constructor args and kwargs directly update the object `__dict__` and thus define any fields given in its args/kwargs. However, I can't exactly test easily since preforking is not supported on Apple silicon chips.

PR #91, opened 2024-12-18
Fixes a bug in the logger for null response bodies

frostyfan109:


Hoid:
I think we can close this PR since the change here isn't needed anymore

PR #93, opened 2025-01-14
Add assignment override model

ptlharit2:
- Adds the backend logic to pull extra time info (assignment override in Canvas) with our LMS sync
- It also adds a new data table and schema for this info (AssignmentOverride)
- It modifies the assignment & student service to use this info in determining the due date for specific student
- modifies the API layer to relay to the JLP/JLS extensions


app/services/assignment_service.py
cnbennett3, line 1: Not sure why line 187 is there. It looks like maybe a cut-n-paste error? Line 184 checks for None and returns an exception. So at line 187, you already know you have an assignment and can just return it.

app/services/text2.txt
cnbennett3, line 1: This file should be removed if not needed.

app/services/assignment_override_service.py
cnbennett3, line 1: Please consider removing lines 4-12 if this code is no longer needed.

cnbennett3 (changes requested):

Hoid (approved):

app/services/assignment_service.py
Hoid: This issue should be resolved now, right @ptlharit2?

app/models/assignment.py
```diff
@@ -25,7 +25,11 @@ class AssignmentModel(Base):
     last_modified_date = Column(DateTime(timezone=True), server_default=func.current_timestamp())
     is_published = Column(Boolean, server_default='f', nullable=False)
     manual_grading = Column(Boolean, server_default='f', nullable=False)
-    
+    assignment_overrides = relationship("AssignmentOverrideModel", cascade="all, delete-orphan", back_populates="assignment")
+    AssignmentModel = relationship("SubmissionModel", cascade="all,delete", back_populates="assignment")
```
frostyfan109: what is this? should be `submission_model` it seems like, but as far as I can tell you aren't using it and if it's not being used there's no need to make the relationship bidirectional

app/models/assignment.py
```diff
@@ -25,7 +25,11 @@ class AssignmentModel(Base):
     last_modified_date = Column(DateTime(timezone=True), server_default=func.current_timestamp())
     is_published = Column(Boolean, server_default='f', nullable=False)
     manual_grading = Column(Boolean, server_default='f', nullable=False)
-    
+    assignment_overrides = relationship("AssignmentOverrideModel", cascade="all, delete-orphan", back_populates="assignment")
```
frostyfan109: is there a reason to be using delete-orphan here? As far as I understand, this option only considers "orphaned" rows as those who are explicitly removed from the parent's relationship (i.e., `my_assignment.assignment_overrides.remove(x)` would orphan `x`). It does not mean orphan in the traditional relational sense where an orphaned row is any row whose associated parent is deleted.

I think this also may mean that cascade delete is not enabled here (since the `all` option specifically excludes delete and delete-orphan), which means you may see a foreign key violation exception thrown when trying to delete an assignment with overrides referencing it.

app/models/assignment_override.py
```diff
@@ -0,0 +1,29 @@
...
+
+    student = relationship(
+        "StudentModel",
+        foreign_keys="AssignmentOverrideModel.student_id",
+        back_populates="assignment_overrides"
+    )
+    assignment = relationship(
+        "AssignmentModel",
+        foreign_keys="AssignmentOverrideModel.assignment_id",
+        backref=backref("assignment_overrides")
```
frostyfan109: why is student using back_populates but assignment is using backref?

app/services/assignment_override_service.py
```diff
@@ -0,0 +1,34 @@
...
+                assignment_override = AssignmentOverrideModel(
+                    id=override["id"],
+                    student_id=student_id,
+                    assignment_id=override["assignment_id"],
+                    available_date=override["unlock_at"],
+                    due_date=override["due_at"]
+                )
+
+                self.session.add(assignment_override)
+                self.session.commit()
```
frostyfan109: Commit should occur a single time after completing iteration. Otherwise, each iteration, you are pushing a single record, committing the transaction (and hoping it succeeds), then repeating.

Not particularly relevant after you change this, but it's also worth keeping in mind that flushing/committing involves network action and flushing specifically involves DB errors such as integrity errors, so these are liable to failure. (Committing implicitly flushes). Which explains in part why committing in sequence in bad (e.g., the first 2 rows get committed but the third fails its flush to the database, resulting in an internal error wherein only those first 2 rows get created and the rest gets aborted without cleanup).

So, if you are ever committing multiple times in succession like here (generally this is bad practice), or if you've done some other actions prior to committing, you need to be mindful about cleanup in case of failure. This isn't obvious in a lot of services where database action occurs since I've been mindful to design the methods to circumvent the need for such cleanup in most cases. It's a specific design choice where we try to complete database actions first since database transactions are easy to reverse and other cleanup is more difficult (or impossible, e.g. in the case of a git commit).

app/models/assignment_override.py
```diff
@@ -0,0 +1,29 @@
...
+    Integer, Interval, DateTime
+)
+from sqlalchemy.schema import UniqueConstraint
+from sqlalchemy.orm import relationship, backref
+from app.database import Base
+
+class AssignmentOverrideModel(Base):
+    __tablename__ = "assignment_override"
+
+    id = Column(Integer, primary_key=True, index=True)
```
frostyfan109, line 12: I'd like to either change our system to use LMS resource ids consistently as our primary keys across all our LMS resource tables or move away from this practice and store the LMS identifier separately such as `lms_id` to be more clear.

app/services/assignment_override_service.py
```diff
@@ -0,0 +1,34 @@
+from sqlalchemy import func
+from sqlalchemy.orm import Session
+from app.models import AssignmentOverrideModel
+from app.enums.assignment_status import AssignmentStatus
+
+class AssignmentOverrideService:
+    def __init__(self, session: Session):
+        self.session = session
+
+    async def create_assignment_override(
```
frostyfan109, line 11: Should probably be named `create_assignment_overrides` (plural) since it's creating a list of overrides it seems, or am I misunderstanding?

app/services/canvas_service.py
```diff
@@ -98,7 +98,7 @@ async def get_course(self):
     
     # returns a dictionary of assignments for a course
     async def get_assignments(self):
-        return await self._get(f"courses/{ settings.CANVAS_COURSE_ID }/assignments")
+        return await self._get(f"courses/{ settings.CANVAS_COURSE_ID }/assignments?include[]=overrides")
```
frostyfan109: Preferred to use `params` dict kwarg here instead of manually doing the query params.

app/services/assignment_service.py
```diff
@@ -33,7 +39,8 @@ async def create_assignment(
         max_attempts: PositiveInt | None,
         available_date: datetime | None,
         due_date: datetime | None,
-        is_published: bool
+        is_published: bool,
+        assignment_override: list | None
```
frostyfan109: the list generic here should be typed

app/services/lms_sync_service.py
```diff
@@ -72,20 +75,28 @@ async def sync_assignments(self):
                     available_date=assignment["unlock_at"],
                     due_date=assignment["due_at"],
                     is_published=assignment["published"],
+                    assignment_overrides=assignment["overrides"],
```
frostyfan109: this argument doesn't exist I don't think? And the corresponding functionality is not implemented either. Definitely needs to be fixed if I'm not misreading; I didn't see any changes to `update_assignment` nor the schema while reviewing.

app/services/lms_sync_service.py
```diff
@@ -72,20 +75,28 @@ async def sync_assignments(self):
...
-                await self.assignment_service.create_assignment(
-                    id=assignment['id'],
-                    name=assignment['name'], 
-                    due_date=assignment['due_at'], 
-                    available_date=assignment['unlock_at'],
-                    directory_path=assignment['name'],
-                    is_published=assignment['published'],
-                    max_attempts=max_attempts
-                )
+                for override in overrides_collection:
```
frostyfan109: This partially stems from my lack of understanding of how extra time is assigned on Canvas's end, but is this not going to create multiple of the same assignment? The code begins by checking if the assignment id exists in the database, but now when it doesn't, it creates that assignment multiple times iterating over each override for the assignment.

It is also using the identical assignment["id"] as the primary key which is obviously going to cause some issues.

I'll also add that creating the override prior to the assignment is probably going to cause integrity issues as the foreign key will be invalid.

app/services/lms_sync_service.py
```diff
@@ -132,6 +143,7 @@ async def sync_students(self):
                 #create a new student
                 print("student doesn't exist", user_info.onyen)
                 await self.student_service.create_student(
+                    id=id,
```
frostyfan109, line 161: Is there a particular reason that you've pivoted the user models to store canvas resource identifiers as the primary keys? They still create an autoincremented sequence in the model code, which implies a contradictory usage of the column. I mentioned this in another comment, but I'm not exactly opposed to this approach. I'd just rather we make a definitive, consistent decision across all "canvas resource" tables. I do think for what it's worth that it would be better practice in the long run to store this id separately in addition to an autoincremented primary key.

app/services/user/student_service.py
```diff
@@ -1,6 +1,6 @@
 from typing import List
 from app.events import dispatch
-from app.models import StudentModel
+from app.models import StudentModel, UserModel
```
frostyfan109: I would be careful about importing UserModel here. It doesn't seem to be used but could lead to a circular dependency. It's the reason a lot of models and services and such are inline-imported within certain methods.

app/services/user/student_service.py
```diff
@@ -94,6 +96,11 @@ async def get_user_by_onyen(self, onyen: str) -> StudentModel:
             raise NotAStudentException()
         return user
     
+    async def get_user_by_id(self, id: int) -> StudentModel:
```
frostyfan109, line 103: This doesn't actually return anything

frostyfan109 (changes requested):
Generally good but still needs a bit of work, particularly with regards to the syncing/assignment service. Some notes that didn't fit into any specific review comments:

It's my understanding the assignment override completely replaces the old, unused extra time implementation. If so, it should be purged from the codebase (we can leave the migrations, but leaving the model, schemas, and service is a bit confusing and redundant).

It seems like maybe the assignment service part not been finished yet? I don't think any of the extra time functionality inside the assignment service is updated, and it seems to still be relying on the old extra time functionality.

app/models/assignment_override.py
frostyfan109, line 1: The extra time model uses a table argument to enforce a unique constraint on the student-assignment association per row: `__table_args__ = (UniqueConstraint("student_id", "assignment_id"),)`. I think this should probably be included here too, but I'm still kind of confused exactly by the nature of assignment overrides in Canvas.

ptlharit2:
The new table and it's relationships were bit off, as you noticed Griffin. Pushed changes to fixed that and you can confirm if you run a sync. Perhaps the relationship doesn't need to be bidirectional, I haven't considered if that will be something we need.

There are more changes I am testing before I push here which essentially use the override to determine the student/assignment deadline rather than extra-time model. Mimics exactly what Canvas does, so once I push those hopefully it should make more sense Griffin - I think you stumbled upon some changes in the student service that I forgot to remove in the push that I meant to follow up on

frostyfan109:
> The new table and it's relationships were bit off, as you noticed Griffin. Pushed changes to fixed that and you can confirm if you run a sync. Perhaps the relationship doesn't need to be bidirectional, I haven't considered if that will be something we need.
> 
> There are more changes I am testing before I push here which essentially use the override to determine the student/assignment deadline rather than extra-time model. Mimics exactly what Canvas does, so once I push those hopefully it should make more sense Griffin - I think you stumbled upon some changes in the student service that I forgot to remove in the push that I meant to follow up on

Let me know once it's ready to rereview and I'll take another look. Should this be changed to a draft for the moment?

app/services/assignment_service.py
ptlharit2: yep this is resolved

app/services/assignment_override_service.py
ptlharit2: It was creating a list of overrides, so you are right. But i've updated it to now just create one override

app/models/assignment.py
ptlharit2: gotcha that makes sense, poorly read the sqlalchemy docs - removed in newer commit.

app/models/assignment.py
ptlharit2: I wasn't using it so removed it

app/services/assignment_override_service.py
ptlharit2: In general I was creating a list of overrides in the create method. Which wasn't sitting right with the rest of the services we had, and how we were syncing. I changed the over all create method to create a single override. You make a good point, and definitely something to avoid. I think the overall change to the method addresses what you are saying here. resolving this, but feel free to open the comment back up if its not resolved

app/services/lms_sync_service.py
ptlharit2: removed in newer commits

app/services/assignment_service.py
ptlharit2: removed in newer commits - assignment service was getting bloated, it was better I thought to create a new service to handle the override logic

app/services/canvas_service.py
ptlharit2: ah yeah good catch, I hadn't checked the _get implementation. fixed it in a newer commit

app/models/assignment_override.py
ptlharit2: I wanted to check with you on the existing extra time implementation you added before the canvas integration. I am not sure we need that, since Canvas handles all the logic of storing different available/due dates for each student and we can use that as source of truth. I am using the override service now to pull the adjusted dates in the assignment service when we get the assignment for a specific student

app/services/lms_sync_service.py
ptlharit2: I think overall we need to either mimic the id's in Canvas so we can fetch the appropriate resource (student, assignment, assignment override, etc.) based on the id. BUT I also see what you are saying, we can just store that info in a seperate column and leave our ids as they are. I agree with you on us deciding one way or the other. For now i've left my implementation around this as is until we decide. Did you have a preference in which direction we take this?

ptlharit2:
@frostyfan109 thank you for the review! I changed some of the glaring in progress things that seemed wonky, and resolved those conversations.

I left 2 things unanswered -
1. using the Canvas resource ids as primary ids in our DB as opposed to storing those in a seperate table. I left a comment for that elsewhere in this PR, we can come to an agreement on one or the other and I can make the appropriate changes
2. I am not sure how we'll use the extra time code you implemented prior to the canvas integration since canvas fundamentally stores this info in a different way - through overrides. I didn't want to remove any of the extra time code without checking with you on if we should keep it for any other reasons

app/services/lms_sync_service.py
ptlharit2: overhauled this whole sync code

cnbennett3 (approved):

ptlharit2:
> Generally good but still needs a bit of work, particularly with regards to the syncing/assignment service. Some notes that didn't fit into any specific review comments:
> 
> It's my understanding the assignment override completely replaces the old, unused extra time implementation. If so, it should be purged from the codebase (we can leave the migrations, but leaving the model, schemas, and service is a bit confusing and redundant).
> 
> It seems like maybe the assignment service part not been finished yet? I don't think any of the extra time functionality inside the assignment service is updated, and it seems to still be relying on the old extra time functionality.

@frostyfan109 I'd like to merge this PR if you don't have any major concerns. I addressed all feedback except deciding on adding canvas Ids as primary keys, or storing in a separate table.

Whichever direction we go, I think that can be added as a follow up change. 

I tested and verified these changes, the way it works is as a student when you do `GET /api/v1/assignments/self` it will give the assignment with dates specific to that student's available_date/due_date set in Canvas's override. If that student does not have an override, it returns the assignment with the original dates

app/services/lms_sync_service.py
frostyfan109: I think it is definitely better practice to maintain two id columns (internal, lms). I can't think of any potential pitfalls of combining them at the moment, but there likely are some lol

src/services/assignment_service.py
```diff
@@ -356,30 +358,23 @@ def _get_extra_time_model(self) -> ExtraTimeModel | None:
...
-            extra_time = self.extra_time_model.extra_time
-        else:
-            deferred_time = timedelta(0)
-            extra_time = timedelta(0)
-
-        # Base due date + have to defer by whatever amount the available date was deferred by + extra time + base extra time
-        return assignment_due_date + deferred_time + extra_time + self.student_model.base_extra_time
+        return assignment_due_date
 
     def _get_is_available(self) -> bool:
```
frostyfan109: I think this and `_get_is_closed` should be deleted, they seem like they were redundant even before these changes. If not, should adjust them to be async and add await to get_adjusted_available_date since those are now async.

It seems the unit tests are the only place where these are being used. They need to be adjusted accordingly.

src/services/assignment_service.py
```diff
@@ -394,12 +389,12 @@ def _get_is_closed(self) -> bool:
         current_timestamp = self.session.scalar(func.current_timestamp())
         return current_timestamp > adjusted_due_date
     
-    def get_assignment_status(self) -> AssignmentStatus:
+    async def get_assignment_status(self) -> AssignmentStatus:
```
frostyfan109: We have a `get_assignment_status` method in both the `StudentAssignmentService` and `InstructorAssignmentService` and they do the same thing (but consume different student vs professor instance method implementations).

It probably makes sense to maintain sync-async parity between both services, e.g., the student implementation here is async now but the instructor implementation is synchronous. I don't particularly think it matters that the instructor version isn't _actually_ doing anything async, it's more a matter of consistency.

src/services/assignment_service.py
```diff
@@ -356,30 +358,23 @@ def _get_extra_time_model(self) -> ExtraTimeModel | None:
...
         return extra_time_model
 
-    # The release date for a specific student, considering extra_time
-    def get_adjusted_available_date(self) -> datetime | None:
-        assignment_open_date = self.assignment_model.available_date or self.course_model.start_at
+    # The release date for a specific student, considering assignment override for this student
+    async def get_adjusted_available_date(self) -> datetime | None:
+        assignment_override = await self.assignment_override_service.get_assignment_override_by_student_id(self.assignment_model.id, self.student_model.id)
+
+        assignment_open_date = (assignment_override.available_date if assignment_override else None) or self.assignment_model.available_date or self.course_model.start_at
```
frostyfan109: Would it make more sense to format this logic as a double ternary?
```python
assignment_open_date = (
  assignment_override.available_date if assignment_override else \
  self.assignment_model.available_date if self.assignment_model.available_date \
  else self.course_model.start_at
)
```
I feel like the logic is a bit more legible this way, even if longer. Same goes for the due date logic.

src/services/assignment_service.py
```diff
@@ -346,6 +347,7 @@ def __init__(self, session: Session, student_model: StudentModel, assignment_mod
         self.assignment_model = assignment_model
         self.course_model = course_model
         self.extra_time_model = self._get_extra_time_model()
```
frostyfan109: I don't think we need to explicitly purge the migrations and models from our codebase for now, but we should go through and remove references throughout the services to extra time, as long as it's not being used anymore. Methods like `get_earliest_available_date` and `get_latest_due_date` aren't actually used in the codebase as far as I know and can probably just be outright deleted.

src/models/assignment_override.py
```diff
@@ -0,0 +1,27 @@
...
+    due_date = Column(DateTime(timezone=True))
+    available_date = Column(DateTime(timezone=True))
+
+    student = relationship(
+        "StudentModel",
+        foreign_keys="AssignmentOverrideModel.student_id"
+    )
+    assignment = relationship(
+        "AssignmentModel",
+        foreign_keys="AssignmentOverrideModel.assignment_id"
```
frostyfan109: I believe both of these should set a backref of "all,delete" instead of an "all" cascade in the assignment model itself. If either of the student or assignment is deleted, it must cascade delete the override or you'll hit a foreign key violation.

Note that cascade="all" explicitly excludes deletion, which is why you need to specify it in addition.

frostyfan109 (commented):
As discussed with Harit, I've had a lot come up in school/personal life which is preventing me from finishing this in a timely manner. I've reviewed all files except the following, which still deserve some attention from another's eye:
- New assignment override service
- LMS service
- assignment_service
- canvas_service
- instructor and student_service

src/services/assignment_override_service.py
```diff
@@ -0,0 +1,77 @@
...
+    ) -> list[AssignmentOverrideModel]:
+        return self.session.query(AssignmentOverrideModel) \
+            .filter_by(assignment_id=assignment_id) \
+            .all()
+    
+    async def get_assignment_override_by_student_id(
+        self,
+        assignment_id: int,
+        student_id: int
+    ) -> AssignmentOverrideModel:
```
frostyfan109: Return annotation should be `| None`, or it should explicitly raise a custom exception like AssignmentOverrideNotFound or something. `queryset.first()` returns None if no rows are returned by the query, but it looks like you're trying to catch this as an error in the lms code.

src/services/lms_sync_service.py
```diff
@@ -75,17 +75,43 @@ async def sync_assignments(self):
                     max_attempts=max_attempts
                 ))
 
+                # Update assignment override if they exist for this assignment in the DB, 
+                # Otherwise create them for new students added to the override
+                if(assignment["has_overrides"]):
```
frostyfan109: style fix: remove parentheses

src/services/lms_sync_service.py
```diff
@@ -75,17 +75,43 @@ async def sync_assignments(self):
...
+                    for canvas_override in assignment["overrides"]:
+                        for student_id in canvas_override["student_ids"]:
+                            try:
+                                db_override = await self.assignment_override_service.get_assignment_override_by_student_id(
+                                    assignment_id=assignment["id"],
+                                    student_id=student_id
+                                )
+                                await self.assignment_override_service.update_assignment_override(db_override, canvas_override)
+                            except:
```
frostyfan109: Ideally annotate the error type here. With the way get_assignment_override_by_student_id currently works, it's going to return `None` for db_override and then feed that into `update_assignment_override` which will cause some sort of attribute error I think; the exception handling here needs some attention.

src/services/lms_sync_service.py
```diff
@@ -75,17 +75,43 @@ async def sync_assignments(self):
...
+                                await self.assignment_override_service.create_assignment_override(canvas_override, student_id)
+                else:
+                    # Delete assignment overrides if they exist for this assignment in the DB, but do not exist in Canvas
+                    for db_override in await self.assignment_override_service.get_assignment_overrides_by_assignment_id(assignment["id"]):
+                        for student_override in db_override:
```
frostyfan109: I'm pretty sure this for loop needs to be deleted. the method above returns a `list[AssignmentOverrideModel]`

src/services/lms_sync_service.py
```diff
@@ -75,17 +75,43 @@ async def sync_assignments(self):
...
+                    name=assignment["name"], 
+                    due_date=assignment["due_at"], 
+                    available_date=assignment["unlock_at"],
+                    directory_path=assignment["name"],
+                    is_published=assignment["published"],
                     max_attempts=max_attempts
                 )
+
+                # Create assignment overrides if the new assignment has one
+                if(assignment["has_overrides"]):
```
frostyfan109: style fix, weird parentheses

src/services/lms_sync_service.py
```diff
@@ -246,7 +274,7 @@ async def upsync_assignment(
     async def downsync(self):
         print("Syncing the LMS with the database")
         await self.sync_course()
-        await self.sync_assignments()
         await self.sync_students()
+        await self.sync_assignments()
```
frostyfan109: I understand this needs to be moved so that students are synced prior to trying to create overrides for them. Can this be done after syncing instructors as well? I think it should be fine. It feels weird to perform this as an intermediate step between user sync steps.

src/services/user/instructor_service.py
```diff
@@ -12,6 +12,7 @@ async def list_instructors(self) -> List[InstructorModel]:
 
     async def create_instructor(
         self,
+        id: int,
```
frostyfan109: this is never used

frostyfan109 (changes requested):
last few files

src/services/assignment_service.py
ptlharit2: I went ahead and deleted those methods. As it stand we have quite a few test suites that are outdated - they haven't been updated since we started actively developing the canvas & gitea integration. We do need to go through every suite and update existing tests, plus add tests for new functionality. I think that should be tracked separately

src/services/assignment_service.py
ptlharit2: yeah agreed, this is more legible - made the update

src/services/assignment_service.py
ptlharit2: I am leaving the migrations and the data models as it. But I deleted the methods and other logic in the assignment_service based on our convo

src/models/assignment_override.py
ptlharit2: good catch, made the update

src/services/assignment_override_service.py
ptlharit2: I added a custom error to throw if no override is found - I am catching that in the sync service. similar flow to how we handle other resources

src/services/lms_sync_service.py
ptlharit2: the override delete method deletes a single instance/row in the database. my thought is we'd want to delete all overrides associated with an assignment (there can be multiple) - so we'd want to loop through all of them in the list and delete them?

Unless `session.delete()` deletes multiple rows? I may need to read the docs to see if that's how it works

src/services/lms_sync_service.py
ptlharit2: yup I added a custom exception to catch when the getter returns a `None`. The exception is caught in the sync service, similar to how we are handling other resources

src/services/lms_sync_service.py
ptlharit2: yeah good point, made the update

PR #94, opened 2025-01-14
Logger updates

Hoid:


app/core/middleware/logger.py
```diff
@@ -143,5 +136,6 @@ async def _execute_request(
     def _get_endpoints_to_not_log_req_body(self) -> list[str]:
         return [
             "/api/v1/login",
+            "/api/v1/login/gitea/ssh",
```
frostyfan109: login/gitea/ssh contains a public key, not private; doesn't need to be omitted

frostyfan109 (approved):

PR #100, opened 2025-04-02
Websocket pubsub merge

frostyfan109:
Fixes the weird refactoring from the poetry merge by just reimplementing the changes. Kept the name as `app` since I did not see the point in renaming to src, and it introduces way too much complexity with the merging


frostyfan109:
did not work.

